PR #126, opened 2015-12-02
Update C# docs for the new auth API.

jtattermusch:
Update C# to the new auth API coming in 0.12 release.
This should be enough to address C# portion of https://github.com/grpc/grpc/issues/4243.

@jboeuf  for sanity check. Besides JWT, I decided to cover also OAuth2 and  per-rpc cases, because they might come handy.

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -122,13 +122,13 @@ stub = Helloworld::Greeter::Stub.new('localhost:50051', creds: creds)
 
 ```csharp
 // Base case - No encryption/authentication
-var channel = new Channel("localhost:50051", Credentials.Insecure);
+var channel = new Channel("localhost:50051", ChannelCredentials.Insecure);
 var client = new Greeter.GreeterClient(channel);
 ...
 
 // With server authentication SSL/TLS
-var credentials = new SslCredentials(File.ReadAllText("ca.pem"));  // Load a CA file
-var channel = new Channel("localhost:50051", credentials);
+var channelCredentials = new SslCredentials(File.ReadAllText("ca.pem"));  // Load a CA file
+var channel = new Channel("localhost:50051", channelCredentials);
 var client = new Greeter.GreeterClient(channel);
 ```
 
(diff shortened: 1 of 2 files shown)


docs/guides/auth.md
```diff
@@ -122,13 +122,13 @@ stub = Helloworld::Greeter::Stub.new('localhost:50051', creds: creds)
...
-var channel = new Channel("localhost:50051", Credentials.Insecure);
+var channel = new Channel("localhost:50051", ChannelCredentials.Insecure);
 var client = new Greeter.GreeterClient(channel);
 ...
 
 // With server authentication SSL/TLS
-var credentials = new SslCredentials(File.ReadAllText("ca.pem"));  // Load a CA file
-var channel = new Channel("localhost:50051", credentials);
+var channelCredentials = new SslCredentials(File.ReadAllText("ca.pem"));  // Load a CA file
+var channel = new Channel("localhost:50051", channelCredentials);
```
jboeuf: I think that you will need an ssl override in this case. Typically, you won't have localhost in your server cert...

The other thing that you could do is show how you would connect to a server with a publicly trusted server (in which case you would not need ca.pem (which would be better named as roots.pem BTW)).

jtattermusch: changed url to a domain name and renamed ca.pem. PTAL.

docs/guides/auth.md
```diff
@@ -122,13 +122,13 @@ stub = Helloworld::Greeter::Stub.new('localhost:50051', creds: creds)
...
 // Base case - No encryption/authentication
-var channel = new Channel("localhost:50051", Credentials.Insecure);
+var channel = new Channel("localhost:50051", ChannelCredentials.Insecure);
 var client = new Greeter.GreeterClient(channel);
 ...
 
 // With server authentication SSL/TLS
-var credentials = new SslCredentials(File.ReadAllText("ca.pem"));  // Load a CA file
-var channel = new Channel("localhost:50051", credentials);
+var channelCredentials = new SslCredentials(File.ReadAllText("roots.pem"));  // Load a CA file
```
jboeuf: Maybe change the comment to:
`// Load a custom roots file.`

jtattermusch: Done.

jboeuf:
LGTM. Thanks much for the fixes. @LisaFC your turn!

docs/guides/auth.md
```diff
@@ -175,26 +175,41 @@ var scope = 'https://www.googleapis.com/auth/grpc-testing';
 
 ```csharp
 // Base case - No encryption/authentication
```
LisaFC: This looks really good - my only suggestion (given that there are now multiple possible approaches for auth with Google) is that for clarity maybe we should split this into multiple code blocks (no auth, recommended approach with JWT access token, legacy approach with OAuth2, and per-RPC) with headings rather than just comments within a big code block.

It's also not super clear how each approach ties in with the generic "Authenticating with Google" paragraph up yonder, though that might be because it also needs some work - who should I prod about doing a separate PR for that?

jtattermusch: Good idea, I've made the splitting. Still not sure if it's 100% clean, but definitely looks much better than before.Ideas what more to adjust are welcome.

jtattermusch: Actually, it might look nice if we had 3 tabs  for the auth example:
by default you would see Auth using JWT (the recommended approach) and you would be able to switch to  OAuth2 auth or single RPC.

But let's leave this improvement for later.

LisaFC: Yeah, I'm happy to merge this now and we can make further improvements in a future PR. Who is the person to talk to for the more "generic" Authenticating With Google paragraph above?

jtattermusch:
Regarding the "Authenticating with Google" paragraph: Let's talk to @jboeuf. If we plan any fundamental changes, it might make sense to loop in @a11r .

PR #129, opened 2015-12-08
Updated auth document with 0.12 Node changes

murgatroid99:
These changes mirror the earlier C# changes


jtattermusch:
@jboeuf  CC

PR #131, opened 2015-12-14
Update PHP auth code example

stanley-cheung:
with the new Call Credentials options.

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -302,7 +302,26 @@ $client = new helloworld\GreeterClient(
   new Grpc\BaseStub('localhost:50051', []));
 ...
 
-// Authenticating with Google
+// Authenticate using JWT access token (recommended approach)
+function updateAuthMetadataCallback($context)
+{
+    $authUri = $context->service_url;
+    $methodName = $context->method_name;
+    $auth_credentials = ApplicationDefaultCredentials::getCredentials();
+
+    return $auth_credentials->updateMetadata($metadata = [], $authUri);
+}
+$opts = [
+  'credentials' => Grpc\Credentials::createSsl(file_get_contents('ca.pem'));
+];
+
+$client = new helloworld\GreeterClient('localhost:50051', $opts);
+$request = new helloworld\HelloRequest();
+$call = $client->SayHello($request, $metadata=[], $options=[
+  'call_credentials_callback' => 'updateAuthMetadataCallback'
+]);
+
+// Authenticate using Oauth2 token (legacy approach)
 // the environment variable "GOOGLE_APPLICATION_CREDENTIALS" needs to be set
 $scope = "https://www.googleapis.com/auth/grpc-testing";
 $auth = Google\Auth\ApplicationDefaultCredentials::getCredentials($scope);
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -307,12 +307,12 @@ function updateAuthMetadataCallback($context)
     $auth_credentials = ApplicationDefaultCredentials::getCredentials();
     return $auth_credentials->updateMetadata($metadata = [], $context->service_url);
 }
-$channelCredentials = Grpc\ChannelCredentials::createComposite(
+$channel_credentials = Grpc\ChannelCredentials::createComposite(
     Grpc\ChannelCredentials::createSsl(file_get_contents('roots.pem')),
     Grpc\CallCredentials::createFromPlugin('updateAuthMetadataCallback')
 );
 $opts = [
-  'credentials' => $channelCredentials
+  'credentials' => $channel_credentials
 ];
 $client = new helloworld\GreeterClient('greeter.googleapis.com', $opts);
 
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -298,39 +298,31 @@ var result = client.SayHello(request, new CallOptions(credentials: googleCredent
 
 ```php
 // Base case - No encryption/authorization
-$client = new helloworld\GreeterClient(
-  new Grpc\BaseStub('localhost:50051', []));
+$client = new helloworld\GreeterClient('localhost:50051', []);
 ...
 
-// Authenticate using JWT access token (recommended approach)
+// Authenticate using scopeless credentials (recommended approach)
 function updateAuthMetadataCallback($context)
 {
-    $authUri = $context->service_url;
-    $methodName = $context->method_name;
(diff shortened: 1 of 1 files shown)


docs/guides/auth.md
```diff
@@ -302,7 +302,26 @@ $client = new helloworld\GreeterClient(
   new Grpc\BaseStub('localhost:50051', []));
 ...
 
-// Authenticating with Google
+// Authenticate using JWT access token (recommended approach)
```
jboeuf: The application default creds may or may not be a JWT access token.

docs/guides/auth.md
```diff
@@ -302,7 +302,26 @@ $client = new helloworld\GreeterClient(
...
+function updateAuthMetadataCallback($context)
```
jboeuf: This looks very manual... Could we have a function that creates a channel credentials using the composite method?

docs/guides/auth.md
```diff
@@ -302,7 +302,26 @@ $client = new helloworld\GreeterClient(
...
+    $methodName = $context->method_name;
+    $auth_credentials = ApplicationDefaultCredentials::getCredentials();
+
+    return $auth_credentials->updateMetadata($metadata = [], $authUri);
+}
+$opts = [
+  'credentials' => Grpc\Credentials::createSsl(file_get_contents('ca.pem'));
+];
+
+$client = new helloworld\GreeterClient('localhost:50051', $opts);
```
jboeuf: 2 issues here:

1. It looks like $opts is kinda optional and that would mean that by default, we create an insecure stub. Am I getting this right?
2. It is unsafe to send google auth tokens to something else that an google endpoint. 'localhost:50051' may be ok since you're sending that locally but:
  - This does not set a good example.
  - You will need some kind of ssl override to make that work.

jboeuf:
Assigning to me since it looks like there are some API level issues. Will reassign to @LisaFC when these are fixed.

docs/guides/auth.md
stanley-cheung: @jboeuf Thanks for the feedback. I lifted this code example from the interop tests so it may not be the best for documentation. How about if I change the example to this?

```
// Authenticate using scopeless credentials (recommended approach)
function updateAuthMetadataCallback($context)
{
    $authUri = $context->service_url;
    $methodName = $context->method_name;
    $auth_credentials = ApplicationDefaultCredentials::getCredentials();

    return $auth_credentials->updateMetadata($metadata = [], $authUri);
}
$credentials = Grpc\ChannelCredentials::createComposite(
    Grpc\Credentials::createSsl(file_get_contents('ca.pem')),
    Grpc\CallCredentials::createFromPlugin('updateAuthMetadataCallback')
);
$opts = [
  'credentials' =>  $credentials
];

$client = new helloworld\GreeterClient('greeter.googleapis.com', $opts);
```

I hope that's cleaner.

jboeuf: Thanks much. That's way better. How about we simplify the approach a little more.
- simplified the callback.
- standardize on camel-case.
- Using Grpc\ChannelCredentials as opposed to Grpc\Credentials (I assume that this is the right call now).
- replaced ca.pem with roots.pem which is our standard publicly trusted roots file (etc/roots.pem).
- changed $credentials to $channelCredentials to make things more explicit.

```php
// Authenticate using scopeless credentials (recommended approach)
function updateAuthMetadataCallback($context)
{
    $authCredentials = ApplicationDefaultCredentials::getCredentials();
    return $authCredentials->updateMetadata($metadata = [], $context->authUri);
}
$channelCredentials = Grpc\ChannelCredentials::createComposite(
    Grpc\ChannelCredentials::createSsl(file_get_contents('roots.pem')),
    Grpc\CallCredentials::createFromPlugin('updateAuthMetadataCallback')
);
$opts = [
  'credentials' =>  $channelCredentials
];

$client = new helloworld\GreeterClient('greeter.googleapis.com', $opts);
```

It would actually be great to have a helper for this code (a function like `createDefaultGoogleChannelCreds` that would return the $channelCredentials) and tell people use this since I think that this will be the 80% use case.

Sorry if my example is not correct BTW, I haven't touched PHP for at last 15 years :).

jboeuf: And we still need to take care of the "Insecure has to be explicit" security requirement. It could be as simple as requiring the $opts parameter in the client and that the $credentials field in $opts to be set (with `Grpc\ChannelCredentials::createInsecure` for insecure).

stanley-cheung: Updated the code example as feedback. Remaining TODO items:
 * ```Grpc\ChannelCredentials::createInsecure``` needs to be explicit
 * create a helper function like ```createDefaultGoogleChannelCreds```

Will address these 2 on the grpc/grpc repo.

jboeuf:
Thanks much for the fixes. Plan sounds great! Please assign the upcoming PRs to me. Once you add the helper function, we should probably use it in this documentation so I'm not sure whether to leave this particular PR open or to merge it now (after @LisaFC approves it) and revisit once we have the helper function in the code. Your choice.

stanley-cheung:
Fixed the underscore case. I'd prefer to merge this now as is and update this again with another PR when the other 2 changes come in.

Thanks for the feedback!

jboeuf:
SGTM. Assigning back to @LisaFC. Thanks for the fixes!

docs/guides/auth.md
```diff
@@ -298,20 +298,31 @@ var result = client.SayHello(request, new CallOptions(credentials: googleCredent
 
 ```php
 // Base case - No encryption/authorization
```
LisaFC: Can you switch these to headings rather than just comments in the snippet, as per the other examples?

LisaFC:
My main concern with this is consistency with the rest of the doc - the other languages refer to JWT tokens, this refers to general scopeless tokens.  Is this because PHP is different or is it also the case in other languages that the non-OAuth2 token isn't necessarily a JWT?

We also need to clarify the general information at the top of the doc, probably in a separate PR (who would be the right person for this?) as the terminology being used right now isn't entirely clear as to which bit of the general info applies to which bit of the examples.

jboeuf:
@LisaFC. Regarding "scopeless" vs "JWT Access", I actually sent you https://github.com/grpc/grpc.github.io/pull/132 yesterday that fixes the issue for all languages (and therefore makes it consistent). Does that work for you?

LisaFC:
Yes, it does, sorry I missed that - that makes much more sense.

PR #132, opened 2015-12-15
Google default credentials may or may not be JWT Access.

jboeuf:
- Changing JWT Access to a more general "Scopeless credentials" for the
  recommended approach to authenticate to Google.


jboeuf:
@jtattermusch @murgatroid99 @stanley-cheung @soltanmm @nathanielmanistaatgoogle please have a look as well. Sorry for not catching that during the C# review.

jtattermusch:
LGTM.

PR #134, opened 2015-12-16
Updated Ruby auth doc with newest Ruby changes

murgatroid99:
Also updated the URLs used in the Node auth examples

CC @jboeuf

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -226,7 +226,7 @@ var ssl_creds = grpc.credentials.createSsl(root_certs);
 (new GoogleAuth()).getApplicationDefault(function(err, auth) {
   var call_creds = grpc.credentials.createFromGoogleCredential(auth);
   var combined_creds = grpc.credentials.combineChannelCredentials(ssl_creds, call_creds);
-  var stub = new helloworld.Greeter('localhost:50051', combined_credentials);
+  var stub = new helloworld.Greeter('greeter.googleapis.com', combined_credentials);
 });
 ```
 
@@ -243,7 +243,7 @@ var scope = 'https://www.googleapis.com/auth/grpc-testing';
   }
   var call_creds = grpc.credentials.createFromGoogleCredential(auth);
   var combined_creds = grpc.credentials.combineChannelCredentials(ssl_creds, call_creds);
-  var stub = new helloworld.Greeter('localhost:50051', combined_credentials);
+  var stub = new helloworld.Greeter('greeter.googleapis.com', combined_credentials);
 });
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guides/auth.md
```diff
@@ -153,12 +153,12 @@ described above appear in each language: more languages are coming soon.
...
 ```ruby
 # Base case - No encryption
-stub = Helloworld::Greeter::Stub.new('localhost:50051')
+stub = Helloworld::Greeter::Stub.new('localhost:50051', :this_channel_is_insecure)
 ...
 
 # With server authentication SSL/TLS
 creds = GRPC::Core::Credentials.new(load_certs)  # load_certs typically loads a CA roots file
-stub = Helloworld::Greeter::Stub.new('localhost:50051', creds: creds)
+stub = Helloworld::Greeter::Stub.new('localhost:50051', creds)
```
jboeuf: please use 'greeter.googleapis.com' as opposed to 'localhost:50051' which would most certainly need an ssl name override.

docs/guides/auth.md
```diff
@@ -180,7 +180,7 @@ var client = new Greeter.GreeterClient(channel);
 ####Base case - No encryption/authentication
```
jboeuf: @LisaFC 
after this PR is merged. I'll do a pass through the doc to remove all the:
 ####Base case - No encryption/authentication
which are under
###Authenticating with Google (language X).

These sections do not  make much sense.

docs/guides/auth.md
murgatroid99: This example actually only exists for Ruby and C#. Do we even need it?

jboeuf: I'm fine removing them.

murgatroid99: Can you do that in the PR you mentioned in the other comment?

jboeuf: wfm.

docs/guides/auth.md
LisaFC: In what sense? Obviously, yes, they aren't actually an example of authenticating with anything but is it not useful to see the difference between "no authentication" and "here's how you add x form of authentication"? Or is it the case that they shouldn't be under Authenticating with Google as they're also not authenticating with anything else?

Perhaps we need to reorganize the whole examples section by language, so then you'd have eg a Ruby section with first "no authentication", then "SSL/TLS authentication", then "authenticating with Google"....

jboeuf: @LisaFC Yes, my plan was to do pretty much what the your last paragraph said and re-organize sections by language.
The subsections should be:
- no authentication cleartext
- server-only authentication with SSL/TLS
- client authentication with Google over server-authenticated SSL/TLS

Sounds good? We could add a section later for mutual authentication with SSL/TLS with SSL client certs but the setup is slightly more complicated and that would just add noise at this point.

LisaFC: Yes, that makes perfect sense!

PR #137, opened 2016-01-08
Update PHP auth example

stanley-cheung:
* Insecure ChannelCredentials is now explicit
 * Update section headings to be consistent with other languages


jboeuf:
LGTM. Thanks much for the fixes. Reassigning to @LisaFC.

PR #138, opened 2016-01-09
Update gRPC Python plugin instructions

awan1:
Changed the link for the gRPC Python plugin instructions to point to https://github.com/grpc/grpc/tree/release-0_11/src/python instead of https://github.com/grpc/grpc/blob/release-0_11/INSTALL; the latter has a much longer and difficult install process.

I'm not familiar with Jekyll so I'm not sure if the site.data.config.grpc variable works properly.

### DIFF
--- a/docs/tutorials/basic/python.md
+++ b/docs/tutorials/basic/python.md
@@ -104,7 +104,7 @@ message Point {
 
 ## Generating client and server code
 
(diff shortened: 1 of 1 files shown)


awan1:
And actually, instead of release-0_11 maybe using master would be better? Right now release 0_11 is outdated.

nathanielmanistaatgoogle:
My apologies but I think the right thing to do with this for now is to sit on it. In the next few weeks we're going to be coming up with a better story for distributing the `protoc` plugin and the content on that page is intended to stay version-locked to whatever version is current.

I'm not denying that we have some complicated and wrong instructions out there, especially for the `protoc` plugin...

awan1:
No problem! That makes sense. I'll just have to figure things out by filing other issues, then :).

PR #146, opened 2016-02-17
fix a 404 url

abop:
fix 404


stanley-cheung:
Dupe of #144. Thanks for the contribution anyway!

PR #150, opened 2016-02-26
Fixed wrong link

johscheuer:
I corrected a link in the [basic tutorial for go](http://www.grpc.io/docs/tutorials/basic/go.html#defining-the-service). 

The old link https://github.com/grpc/grpc-go/blob/master/examples/helloworld/proto/route_guide.proto resulted in a 404 error.


LisaFC:
Already fixed in https://github.com/grpc/grpc.github.io/pull/165

PR #156, opened 2016-03-11
Fix C# non-Windows `protoc` plugin argument

KevM:
Trying to do this myself it failed. I noticed that the Windows command included the plugin name before the location of the plugin executable.


jtattermusch:
Thanks for pointing this.

PR #158, opened 2016-03-15
fix wrong .proto file path

favadi:


favadi:
Close, duplicate with #150 .

PR #195, opened 2016-05-13
Get rid of the outdated installation directory

jtattermusch:
-- the docs in this directory in basically an outdated copy of the language-specific README.md files under https://github.com/grpc/grpc/tree/master/examples
-- the docs is out of date and is not linked to from anywhere.
-- we really want to dedupe.


LisaFC:
Yup, good idea.

PR #197, opened 2016-05-24
update tutorial with static/dynamic info

LisaFC:
Updated tutorial to reflect the new static/dynamic code generation options in Node. Will need to wait til the next release to merge though as the links are all to the most recent release, which doesn't have the new examples... (they're in master)

### DIFF
--- a/_data/config.yml
+++ b/_data/config.yml
@@ -1,2 +1,2 @@
-grpc_release_branch: release-0_13
+grpc_release_branch: release-0_14
 milestones_link: https://github.com/grpc/grpc/milestones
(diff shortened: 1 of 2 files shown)


murgatroid99:
The relevant release has been published and the examples are there: https://github.com/grpc/grpc/tree/release-0_14/examples/node

Otherwise, the changes LGTM

LisaFC:
Ah, ok, I think the variable the site is using in links is still sending everything to the 0_13 release, should we change that?

murgatroid99:
I think that would be a good thing to change.

PR #203, opened 2016-06-02
Updated installation instructions

LisaFC:
(partly?) fixes https://github.com/grpc/grpc.github.io/issues/202


hsaliak:
LGTM, merging so that we can iteratively introduce more changes.

PR #206, opened 2016-06-10
Google authentication cleanup.

### LINKED ISSUE #180: Remove deprecated scoped credentials examples from auth.md
This approach is described in the Examples section as no longer recommended and, per @jboeuf, the snippets just add "noise" to the doc.

jboeuf:
- Removing legacy way to authenticate with Google.
- Java and Python will need to update their examples to use "scopeless"
  creds.
- Fixes #180.

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -169,26 +169,13 @@ creds = GRPC::Core::Credentials.new(load_certs)  # load_certs typically loads a 
 stub = Helloworld::Greeter::Stub.new('myservice.example.com', creds)
 ```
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
An issue for python to update the example is here: https://github.com/grpc/grpc/issues/6860. Also, as per that issue, python JWT is not going to be part of GA.

jtattermusch:
Changes look good to me, unassigning myself and waiting for @LisaFC  to approve.

jtattermusch:
Actually unassigning myself now.

PR #217, opened 2016-06-29
Added "release" URL

dgquintas:
In order to get the branch name for the current release. This will be used throughout tutorials and examples where checking out a specific branch is better than working at HEAD. For example

```
$ git clone -b $(curl http://grpc.io/release) github.com/grpc/grpc
```

### DIFF
--- a/release
+++ b/release
@@ -1,0 +1,5 @@
+---
+permalink: /release
+---
+
+{{ site.data.config.grpc_release_branch }}


dgquintas:
This change will be used in addressing https://github.com/grpc/grpc/issues/3899

hsaliak:
As discussed offline, please make jekyll ignore this file so it does not render it as html.

hsaliak:
LGTM

Issue #181, opened 2016-04-26
Add remaining language examples to auth.md


LisaFC:
Currently we're missing (as far as I can see): 

- some C++ examples (I know there are some in the general API section but they should be comprehensive and in Examples as well, as that's where users are going to look for them)
- some Objective C examples (there are none)
- just using SSL/TLS example for PHP (it looks like it's in the auth with Google example but I don't know enough about PHP to pull it out separately)
- Scopeless credentials example for Java (if it exists)

LisaFC:
Ping?

jboeuf:
#204 addresses C++, assigning language owners for other languages.

ejona86:
Scopeless credentials example needs grpc/grpc-java#785, and then would be similar to other languages where you simply don't specify credentials.

LisaFC:
Objective C is actually covered in its own tutorial, so can link to it.

PR #222, opened 2016-06-30
Swap Java auth to scopeless credentials

ejona86:
For #181

This requires grpc-java 0.15, which is not yet released.


ejona86:
grpc-java 0.15 is now released, so this is safe to merge.

PR #240, opened 2016-07-14
[Work in Progress] Merges gh-pages into GA branch

yowakita:


yowakita:
I will be submitting a new clean PR with a few more tweaks/clean up.

PR #241, opened 2016-07-14
Merges Wildebeest changes

yowakita:


CNAME
```diff
@@ -1 +0,0 @@
-www.grpc.io
```
hsaliak: Is this intentional?

Gemfile
```diff
@@ -1,2 +1,2 @@
 source 'https://rubygems.org'
-gem 'github-pages'
+gem 'github-pages'
```
hsaliak: Looks like this change is redundant.

community/index.html
```diff
@@ -0,0 +1,68 @@
...
+                        <div class="col-sm-4 contribute-item">
+                            <div class="item-content">
+                                <a class="item-box" href="https://github.com/grpc/grpc-go/labels/help%20wanted">
+                                    <p class="item-link">gRPC Go</p>
+                                </a>
+                                <p class="item-desc">For the Go implementation</p>
+                            </div>
+                        </div>
+                    </div>
+                    <a href="http://wildebeestdev.github.io/grpc.github.io/contribute"><span>More on how to contribute to gRPC Documentation ></span></a>
```
hsaliak: hardcoded link, could you please fix?

contribute/index.html
```diff
@@ -1,39 +1,33 @@
...
+                    <h2>Edit our site on github</h2>
 
-<h3>IRC Channel</h3>
-<p class="lead">Join us on the <a href="http://webchat.freenode.net/?channels=grpc">#grpc IRC channel on Freenode</a> to talk to other developers working on the project.</p>
+                    <p>Click the below button to visit the repo for our site. You can then click the "Fork" button in the upper-right area of the screen to create a copy of our site on your GitHub account called a "fork." Make any changes you want in your fork, and when you are ready to send those changes to us, go to the index page for your fork and click "New Pull Request" to let us know about it.</p>
 
-<h3>Mailing list</h3>
-<p class="lead">Join us on <a href="https://groups.google.com/forum/#!forum/grpc-io">Google Groups</a>. Email: <a href="mailto:grpc-io@googlegroups.com" itemprop="email">grpc-io@googlegroups.com</a></p>
\ No newline at end of file
+                    <button class="btn btn-grpc waves-effect waves-light"><a href="https://github.com/wildebeestdev/grpc.github.io/">Browse this site's source code</a></button>
```
hsaliak: hardcoded link, please fix.

docs/guides/concepts.md
```diff
@@ -1,13 +1,16 @@
...
+headline: gRPC Concepts
 layout: docs
+sidenav: doc-side-guides-nav.html
 title: gRPC Concepts
+type: markdown
 ---
+<p class="lead">This document introduces some key gRPC concepts with an overview of gRPC's architecture and RPC life cycle.</p> 
 
-<h2 class="page-header">gRPC Concepts</h2>
+It assumes that you've read the [Overview]({{site.baseurl}}/docs). For language-specific details, see the Quick Start, tutorial, and reference documentation for your chosen language(s), where available (complete reference docs are coming soon).
```
hsaliak: Does it make sense to associate the right links with the words Quick Start, tutorial and reference documentation?

docs/index.html
```diff
@@ -0,0 +1,126 @@
...
+  </div>
+</div>
+  <hr class="section-divider"/>
+<div class="container">
+  <div class="row">
+    <div class="col-md-11 nofloat center-block">
+      <h2 class="col-sm-8 doc-header">gRPC by Use Cases</h2>
+      <h5 class="col-sm-8 ">gRPC for Last Mile of Computing</h5>
+      <div class="col-sm-10 use-description">
+        <p>gRPC helps generate libraries for iOS and Android as well as offers a JS library for use in browsers.  We are eager to see how it plays in IOT environment as well :-)</p>
```
hsaliak: @zinuga could you suggest revised wording here? IMO we should not talk about browsers in this section yet.

js/common.js
```diff
@@ -20,7 +101,132 @@ function setupTabs(rootElement) {
...
+    $('.toggle').on('click',function(){
+      $(this).toggleClass('active');
+    });
+
+    var forwarding = window.location.hash.replace("#","");
+    if (forwarding) {
+        $("#generalInstructions").hide();
+        $("#continueEdit").show();
+        $("#continueEditButton").text("Edit " + forwarding);
+        $("#continueEditButton").attr("href", "https://github.com/wildebeestdev/grpc.github.io/edit/gh-pages/" + forwarding)
```
hsaliak: hardcoded link, needs to be changed here.

_config.yml
```diff
@@ -1,5 +1,19 @@
 kramdown:
   input: GFM
   hard_wrap: false
+
+markdown:
+  kramdown
+
+baseurl: /grpc.github.io
+
+githuburl: https://github.com/wildebeestdev/grpc.github.io/edit/gh-pages
```
hsaliak: hardcoded link to wildebeest repo needs to be changed.

sitemap.xml
```diff
@@ -0,0 +1,20 @@
...
+    {% endfor %}
+
+    {% for page in site.pages %}
+    {% if page.sitemap_exclude != 'y' %}
+    <url>
+        <loc>{{site.baseurl}}{{ page.url | remove: 'index.html' }}</loc>
+    </url>
+    {% endif %}
+    {% endfor %}
+</urlset>
```
hsaliak: newline at end of file

hsaliak:
Added a few comments, could you please take a look?

@zinuga the favicon only has one arrow, I feel that the short icon, 'G' might be more symmetric with two arrows on either side. wdyt?

docs/guides/concepts.md
LisaFC: I think it's OK "as is" - as they're all language-specific the only place we could send them would be to the Quickstarts, Tutorials, and Reference landing pages, where they'd need to find their language anyway.

_sass/base/_markdown.scss
```diff
@@ -0,0 +1,92 @@
...
+  li {
+    line-height: 24px;
+  }
+
+  // Lists
+  ol,
+  ul {
+    list-style: initial;
+    list-style-position: inside;
+    color: $black;
```
LisaFC: Is there a reason why list items are black while the rest of the text is in text colour (dark grey)? It looks a bit odd/jumpy.

_sass/base/_override.scss
```diff
@@ -0,0 +1,96 @@
...
+  background-color: $thirdBrandColor !important;
+}
+
+code {
+  color: $codeColor;
+  background-color: $inverseBrandColor;
+  font-size: 14px;
+}
+
+pre, pre.prettyprint {
```
LisaFC: Just based on your local Wildebeest preview, pre elements (and, I think, code elements) are rendering in normal text font rather than code font/Courier (it took me a while to figure out why they looked odd....). I know I didn't have to specify it in my original CSS but it seems to be no longer the default...

_sass/base/_markdown.scss
```diff
@@ -0,0 +1,92 @@
...
+  h4 {
+    margin: 20px 25px;
+    font-size: 16px;
+    font-weight: 300;
+    color: $dark-gray;
+    
+    @media (max-width: 768px) {
+      margin: 20px 0;
+    } 
+    // Text immediately following h6 should have same level of indention
```
LisaFC: This looks a bit strange in preview, largely because the indentation is at the same level as code blocks so if there's a code block after the first paragraph (as there is for almost all the h4s I've checked) it looks like a huge chunk of text has inexplicably moved sideways - eg http://wildebeestdev.github.io/grpc.github.io/docs/tutorials/basic/java.html#calling-service-methods

docs/index.html
zinuga: I am just looking to collect interested candidates, so when we launch we know who to reach out to first.

CNAME
yowakita: @hsaliak Sorry that was unintentional, we had to take that out for development but it will be put back in there in a PR.

_config.yml
yowakita: Thanks @hsaliak this will be fixed in a PR shortly

sitemap.xml
yowakita: Thanks this will be fixed in an PR shortly. I'll also add newline to robots.txt

_sass/base/_markdown.scss
yowakita: Thanks @LisaFC I will adjust this to dark-gray

_sass/base/_override.scss
yowakita: Thanks @LisaFC , our reset.scss config took out the default styling of the code and pre elements, so I took them out of the reset and will have it pushed up with a PR shortly.

_sass/base/_markdown.scss
yowakita: @LisaFC could you elaborate a little about how this should look? Would you prefer that the code blocks and its lead are not on the same level? In this screenshot, should the 'Simple RPC' be on the same level of indention as 'Now let’s look at how we call our service methods.' , or should the code block be indented one level further?


screen shot 2016-07-18 at 2 28 34 pm

js/common.js
yowakita: Thanks this will be fixed in a PR shortly

community/index.html
yowakita: Thanks this will be fixed in a PR shortly

_sass/base/_markdown.scss
LisaFC: I'm just not sure that indenting the H4 follow up paragraphs quite works as a whole - in a few examples, the text *before* the H4 is also indented (either a list or a code block), so the effect is more of random indentation rather than highlighting the new heading. Eg http://wildebeestdev.github.io/grpc.github.io/docs/tutorials/basic/java.html#server-side-streaming-rpc  

So yes, in your example I'd have "Simple RPC" and "Calling the simple..." on the same level as "Now let's look at how to..."

Thanks!

yowakita: Thanks for the clarification @LisaFC . How does this look? H4 and their follow up paragraphs are now inline with other text elements. With the change, I believe there would be two levels of indention - One for most html elements, and one level of indention deeper for lists and code blocks

screen shot 2016-07-18 at 3 40 05 pm

LisaFC: I think that's much clearer - thanks!

hsaliak:
LGTM - thanks for making the changes, I will merge this in.

PR #244, opened 2016-07-18
Expanded protocol buffers info in intro

LisaFC:
Based on user feedback, users would like a bit more information about protocol buffers in the actual intro rather than having to go off to the protocol buffers docs. Does this seem like enough extra material?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -61,18 +61,72 @@ your applications.
 <a name="protocolbuffers"></a>
 By default gRPC uses *protocol buffers*, Google’s
 mature open source mechanism for serializing structured data (although it
-can be used with other data formats such as JSON). As you'll
-see in our example below, you define gRPC services using *proto files*,
-with method parameters and return types specified as protocol buffer message
-types. You
-can find out lots more about protocol buffers in the [Protocol Buffers
-documentation](https://developers.google.com/protocol-buffers/docs/overview).
+can be used with other data formats such as JSON). Here's a quick intro to how
+it works. If you're already familiar with protocol buffers, feel free to skip
+ahead to the next section.
+
+The first step when working with protocol buffers is to define the structure
+forthe data you want to serialize in a *proto file*: this is an ordinary text
+file with a `.proto` extension. Protocol buffer data is structured as
+*messages*, where each message is a small logical record of information
+containing a series of name-value pairs called *fields*. Here's a simple
+example:
+
+```
+message Person {
+  string name = 1;
+  int32 id = 2;
+  bool has_ponycopter = 3;
+}
+```
+
+Then, once you've specified your data structures, you use the protocol buffer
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -61,18 +61,72 @@ your applications.
...
-with method parameters and return types specified as protocol buffer message
-types. You
-can find out lots more about protocol buffers in the [Protocol Buffers
-documentation](https://developers.google.com/protocol-buffers/docs/overview).
+can be used with other data formats such as JSON). Here's a quick intro to how
+it works. If you're already familiar with protocol buffers, feel free to skip
+ahead to the next section.
+
+The first step when working with protocol buffers is to define the structure
+forthe data you want to serialize in a *proto file*: this is an ordinary text
```
hsaliak: s/forthe/for the/

docs/index.md
```diff
@@ -61,18 +61,72 @@ your applications.
...
+  string name = 1;
+}
+
+// The response message containing the greetings
+message HelloReply {
+  string message = 1;
+}
+```
+
+With gRPC you also use `protoc` (this time with a special gRPC plugin) to
```
hsaliak: I would change this to something like, 
gRPC uses the `protoc` (..) with a gRPC plugin.. 
The reason is that in some languages, users do not directly interact with the protoc plugin to generate code. For example, in java, the gradle or maven plugin does this for them, similarly for C#. For python, we have a python module that can be invoked. Therefore the language here might be better if it did not imply how the user would user the tool, but how the plugin and protoc tool work together to generate the code.

docs/index.md
```diff
@@ -61,18 +61,72 @@ your applications.
...
+documentation](https://developers.google.com/protocol-buffers/docs/overview),
+and find out how to get and install `protoc` with gRPC plugins in your chosen
+language's Quickstart.
 
 
 #### Protocol buffer versions
 
 While protocol buffers have been available for open source users for some
-time, our examples use a new flavor of protocol buffers called proto3,
+time, our examples use a newer flavor of protocol buffers called proto3,
```
hsaliak: s/newer/latest ? Just a suggestion.

docs/index.md
LisaFC: Well spotted, done.

docs/index.md
LisaFC: Good suggestion, done.

docs/index.md
LisaFC: Done

hsaliak:
really like this and I think users would find this very practical and useful. Would you please be able to repurpose this pull request against the GA branch? I can then merge it once I get in.

LisaFC:
Argh, sorry, this managed to fall between the cracks of my giant to-do list. I'll redo this PR and we can get it in.

LisaFC:
Oh, hang on, it's already in, presumably from when I actually did it. Closing.

PR #247, opened 2016-07-20
Remove extraneous anchor tag that breaks rendering

michaeledgar:
The "Deadlines" header is not rendered correctly on the live [grpc.io site](http://www.grpc.io/docs/guides/concepts.html#deadlines). This is almost certainly due to the manually-inserted anchor tag immediately preceding it.


michaeledgar:
I'll re-send this pull request with a properly created commit with my @google.com email address. (I used the "edit file" UI which used a noreply email instead)

PR #248, opened 2016-07-20
Remove extraneous anchor tag that breaks rendering

michaeledgar:
The "Deadlines" header is not rendered correctly on the live grpc.io
site [0]. This is almost certainly due to the manually-inserted anchor
tag immediately preceding it.

Since the anchor tag in question matches the heading text, it can be
removed entirely rather than moved to the start of the following
paragraph. The fixed heading will have its own automatically-generated
anchor tag with the same name.

[0] http://www.grpc.io/docs/guides/concepts.html#deadlines


hsaliak:
Thanks for the patch! LGTM

PR #252, opened 2016-07-23
The first draft of Objective-C Quickstart

muxi:
The very first draft; should be a lot of revisions needed. Please feel free to provide any comment.

### DIFF
--- a/docs/quickstarts/objective-c-quickstart.md
+++ b/docs/quickstarts/objective-c-quickstart.md
@@ -11,69 +11,99 @@ title: Objective-C Quickstart
 
 ## Before you begin
 
+### System requirement
+OS X El Capitan (version 10.11) or above is required to build and run this Quickstart.
+
 ### Prerequisites
 
 * `CocoaPods`: version 1.0 or higher
-* `XCode`: version 7.2 or higher
+
+   * Check status and version of CocoaPods on your system with command `pod --version`.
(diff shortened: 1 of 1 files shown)


zinuga:
@muxi a few comments
can we tell users how they check if they have cocoapods, xcode, homebrew and what versions they have. 
Also, can we introduce some failure scenarios and how to handle them. 
e.g for me brew install --with-plugins grpc fails with this message.

/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/open-uri.rb:353:in `open_http': 422 Unprocessable Entity (GitHub::Error)
Validation Failed

How to help users proceed in case of such installation errors.

jcanizales:
@muxi, I have a loaner macbook that I got a while ago for the purpose of testing the installation of gRPC on a "fresh" machine. Do you want it?

zinuga:
Just discussed with @muxi  that extending 'hello world' is an important step for a user (adding to proto file, regenerating code, implementing server etc) before actual use of grpc. It is strictly not a part of grpc quickstart but we should include it here as its a critical step before any user uses grpc in their project. So our quickstarts should include running 'helloworld' and extending 'helloworld'.

muxi:
Just tested this new version of Quickstart on a clean Mac OS. I will update with another Troubleshoot section containing common problems and solutions tomorrow morning.

zinuga:
@muxi can you point your PR to GA branch and then we can merge.

PR #254, opened 2016-07-25
Update java guide/tutorial to v0.15.0

dapengzhang0:
- Interfaces in the proto generated code are deprecated, should not be in the example.
- static bindService method is deprecated, should not be in the example.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -269,14 +269,14 @@ To install the gRPC runtime for your chosen language:
 <p>To install the runtime for your own projects, add the following to your build files:</p>
 
 <p>Gradle:</p>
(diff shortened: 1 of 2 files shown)


ejona86:
This LGTM. I'm not sure if it will cause conflicts with #241

dapengzhang0:
One of the files `docs/index.md` will cause conflicts as it is moved and modified in #241.
What is the timeline for #241? @yowakita @hsaliak

hsaliak:
Lets merge this into the master, but we need to follow up and ensure that the changes are also in the GA branch and reflects GA (1.0) than 0.15.0..

PR #265, opened 2016-07-29
Local build \w bundle: update async cpp example be compatible \w latest Liquid

bzz:
This is one of the steps in order to get local bundle workflow working:  
```
bundle install
bundle exec jekyll serve --watch
```

On local env Jekyll 2.4.0 is installed though Bundler (see Gemfile.lock) and by default it fails the build with:

```
Error: Variable '{{' was not properly terminated with regexp: /\}\}/
```

The reason is a known feature in the new Liquid template parser, wich is used since Jekyll 2.0
https://jekyllrb.com/docs/troubleshooting/#liquid - it requires closing
`{{` on the same line `}}`.

This results in ugly long lines, but at least makes Jekyll 2.4.0 build the site locally well.

### DIFF
--- a/docs/tutorials/async/helloasync-cpp.md
+++ b/docs/tutorials/async/helloasync-cpp.md
@@ -8,12 +8,10 @@ This tutorial shows you how to write a simple server and client in C++ using
 gRPC's asynchronous/non-blocking APIs. It assumes you are already familiar with
 writing simple synchronous gRPC code, as described in [gRPC Basics:
 C++](/docs/tutorials/basic/c.html). The example used in this tutorial follows on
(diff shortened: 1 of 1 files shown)


hsaliak:
Thank you so much for this pull request. Would you mind testing this against the 'GA' branch, ditto for the other requests as well. We are overhauling the site for the upcoming release and will switch over to that, so it would make things easier if pull requests went in there directly at least until we switch over.

bzz:
Thanks for the heads up, I did not know about GA. Will be happy to look into it, later this week.

bzz:
Closing as this has been taken care of under 91f717b2eb7eda899def61225386f7cd3a097f50

PR #266, opened 2016-07-29
Fix markdown rendering in Python tutorial

bzz:
Remove extra tag from markdown that breaks rendering of Heading 2. 
Same as #248 but for the Python tutorial.

**Before:**
screen shot 2016-07-29 at 14 39 18

**After:**
screen shot 2016-07-29 at 14 39 37


nathanielmanistaatgoogle:
I'm not sure how this may or may not conflict with a larger change that's happening to this documentation. @LisaFC will know.

LisaFC:
It's fine, it's the new overly finicky Markdown engine that Jekyll now uses, which will be the same after the larger changes.

Issue #237, opened 2016-07-08
Refresh Quickstart content for website. 


hsaliak:
- [x] Volunteers for every language. 
- [x] Review content, provide PRs if necessary.

LisaFC:
We now have a Python quickstart in this PR that should be used as a model for the other languages: https://github.com/grpc/grpc.github.io/pull/223

Preview here: https://github.com/LisaFC/grpc.github.io/blob/quickstarts/docs/quickstarts/python-quickstart.md

zinuga:
- [ ] @murgatroid99 can you do node and ruby?
- [ ] @nicolasnoble C++?
- [ ] @ejona86 java?

Lets follow the python template and get this done

zinuga:
- [ ] @apolcyn for C#

ejona86:
Does the quickstart replace the tutorial? Or is it in addition?

LisaFC:
It's in addition - the idea is that it's for someone who wants to dive straight in and DO SOMETHING, without much explanation, just letting them complete a proper task. Then they look at the tutorial for a walk through with all the context information and links.

I have shamelessly stolen this idea from cloud.google.com's doc template and UX research...

murgatroid99:
Node is unusual here because it has two different kinds of code generation. Dynamic code generation is simpler/easier to use, but static code generation has a workflow and interface more consistent with how this will work in other languages. Which should my Node quickstart guide cover?

hsaliak:
Do we want users to eventually move to the static flow, if so we should
just cover that here so that we do not have to come back and update the
guide later.

If we want to leave the static code generation flow as an advanced feature
and let most users just do dynamic code generation, then i think the latter
is sufficient.

murgatroid99:
I think that if we want the reader to get something up and running quickly and relatively easily, dynamic codegen works better for that. I'm not exactly sure if we want all users to switch to static codegen, but I don't think we intend to remove the dynamic codegen anytime soon. Plus, switching won't be a ton of work, so, I'll go with dynamic codegen for now.

murgatroid99:
Also, should these quickstart guides go into the GA branch or the master branch?

hsaliak:
They should go into the GA branch.

apolcyn:
C# draft here https://github.com/grpc/grpc/pull/7515. Greeter and route_guide examples seemed to need to be updated too, with https://github.com/grpc/grpc/pull/7514.

zinuga:
[...]

PR #269, opened 2016-07-30
Fix quickstart formatting

hsaliak:
Changes needed to display quickstarts correctly. 
Tested on local jekyll instance, confirmed that syntax highlighting works as expected and side-navigation/justification works. 

Completes changes related to #237 for C++.
/cc @LisaFC

### DIFF
--- a/docs/quickstart/cpp.md
+++ b/docs/quickstart/cpp.md
@@ -1,6 +1,9 @@
 ---
+bodyclass: docs
 layout: docs
-title: C++ Quickstart
+headline: C++ Quickstart
+sidenav: doc-side-quickstart-nav.html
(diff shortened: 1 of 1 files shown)


dgquintas:
Bummer, we don't get C++ syntax highlighting. But we do for proto... perhaps the keyword for enabling C++'s is something else in the new system? cpp or something like that?

hsaliak:
On the rendered site, it turns out that i do get color without the explicit cpp markup.. 
I did try some other rendering markup like cpp as you suggested, but then realized that not specifying it was good enough.

image

dgquintas:
LGTM. Merge at will.

PR #273, opened 2016-08-01
Don't line-wrap code snippets

ejona86:
Line-wrapped code looks horrible; just displaying a scroll bar is much
better when possible.

Headers could previously overlap with the scroll bar of code which would
prevent scrolling. This was to create a larger clickable area for
headers. However, the clickable area was much larger than it needed to
be, and simply reducing the area produces apparently-identical results
without breaking scrolling for code snippets.


hsaliak:
LGTM

PR #280, opened 2016-08-02
Update Java tutorial

ejona86:
This synchronizes the tutorial with the current state of the examples
and API (since we recently removed some parts of the API). It also
reduces the indentation of code in code blocks to better utilize the
very small column available.

### DIFF
--- a/_data/config.yml
+++ b/_data/config.yml
@@ -1,2 +1,3 @@
 grpc_release_branch: release-0_14
+grpc_java_release_tag: v1.0.0-pre1
 milestones_link: https://github.com/grpc/grpc/milestones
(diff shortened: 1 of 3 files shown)


_data/config.yml
```diff
@@ -1,2 +1,3 @@
 grpc_release_branch: release-0_14
+grpc_java_release_tag: v1.0.0-pre1
```
hsaliak: Do you want to just flip this to v1.0.0 since this branch will only be live when 1.0 is?

hsaliak:
LGTM please merge as appropriate. The only comment I left is an optional suggestion.

_data/config.yml
ejona86: I considered it. You can see I used to have v1.0.0 hard-coded. I figured making it point to pre1 since now it is easy to bump the version, and it would help to have something copy/pasteable, and grpc_release_branch will also need to be bumped.

hsaliak: Fair enough. This would also be a good test that ensures that hard coded references to 1.0 now point to this variable instead.

PR #286, opened 2016-08-04
Adding meta tag for domain verification. Need for set up of emails.

zinuga:
PR to add metatag to homepage for domain verification


ejona86:
@zinuga LGTM. It's a shame there isn't an easy way to only put it on index.html.

zinuga:
I know. Merging to test domain verification

PR #288, opened 2016-08-04
rename csharp quickstart and move its directory

apolcyn:
This moves and renames the csharp quickstart doc that is currently merged (not the pending one, which is waiting on other PRs)


ejona86:
@apolcyn LGTM. Should I merge?

apolcyn:
@ejona86 this should be safe to merge, only renames the files

PR #289, opened 2016-08-04
Remove unnecessary indentation in code snippets

ejona86:
The new website doesn't have as much vertical space. We don't want to
waste that space on avoidable indentation.

Note for reviewers: this change only includes whitespace changes, as verified with `git diff -w`

### DIFF
--- a/docs/quickstart/node.md
+++ b/docs/quickstart/node.md
@@ -21,12 +21,12 @@ type: markdown
 You'll need a local copy of the example code to work through this quickstart. Download the example code from our GitHub repository (the following command clones the entire repository, but you just need the examples for this quickstart and other tutorials):
 
 ```sh
-  $ # Clone the repository to get the example code
-  $ git clone https://github.com/grpc/grpc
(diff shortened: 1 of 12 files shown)


dgquintas:
Ugh, we should have acquired a lock, I was midway doing this. Thanks!

dgquintas:
Ah, this is only for the code whitespace. I'll take care of the single-line paragraphs then :)

ejona86:
> Ugh, we should have acquired a lock, I was midway doing this. Thanks!

_I_ did. I updated the comment to have `@ejona86` in it (which I expect you don't have enough permission to do :-P).

dgquintas:
hmm those poll-based mechanisms...

ejona86:
> hmm those poll-based mechanisms...

@dgquintas, like a spin lock!

dgquintas:
LGTM, merge at will.

dgquintas:
Actually, it's been a while, merging it myself.

PR #295, opened 2016-08-09
Fixed broken {{ ... }} lines.

dgquintas:


hsaliak:
LGTM 
Verified on a running instance. It does not help that jekyll needs {{ and }} on the same line and the code review for Github line wraps.

PR #306, opened 2016-08-11
update basics with link to quickstart for ide issues

apolcyn:
Left out this basics doc earlier, I think its short instructions on building the example at the end should refer to the quickstart instructions because of the issues with nuget, Xamarin Studio, and Monodevelop.


hsaliak:
LGTM

PR #311, opened 2016-08-12
addressed feedback from fixit 2

dgquintas:


hsaliak:
That! Looks Good. To me.

PR #314, opened 2016-08-12
Fix file path of helloworld.proto

murgatroid99:
The directory in the grpc repo is `examples/protos`, not `examples/proto`.


hsaliak:
Thanks for catching this! LGTM

PR #323, opened 2016-08-16
Fixing supported platforms, adding Juniper reference and hiding coupl…

zinuga:
Adding Java and Go to supported platforms
Adding a caveat for PHP (not 1.0 yet)
Adding a quote from Juniper
Hiding vendasta and mobile benchmarking blogs which we will publish after 1.0
removing mailchimp reference. will implement email subscription after 1.0


hsaliak:
LGTM

PR #329, opened 2016-08-18
Making some more changes to docs and about

zinuga:
Fixing:
about: PHP caveat
docs: gRPC other projects
faq: adding sign up for web etc


hsaliak:
LGTM

Issue #330, opened 2016-08-18
Colorize all logos


hsaliak:
Please colorize all logos from customers.

zinuga:
Done.

PR #345, opened 2016-08-20
Fixing logos in about

zinuga:
Fixing logos for quotes
Italicizing note for proto PHP in supported platforms.


hsaliak:
Fixes #330 
LGTM

PR #347, opened 2016-08-22
Add android as a language in the docs section.

hsaliak:
Fixes #344  Tested locally.

PR #349, opened 2016-08-22
adding android java to the main page

hsaliak:
Add android java to the main page, follow up to #347 to ensure consistency. 
I also re-ordered the list of languages so that both pages are consistent. 
/cc @a11r


zinuga:
LGTM

PR #360, opened 2016-08-24
Remove duplicated FAQ entry

icecrime:
The question "Can I use it in the browser?" appears twice with similar
answers in the FAQ page. Remove one occurrence by keeping what appears
to be the most complete answer.

See the preserved question [here](https://github.com/icecrime/grpc.github.io/blob/59de19cac272024368da21dd8f417f0581a2db8f/faq/index.html#L94-L99).

Signed-off-by: Arnaud Porterie (icecrime) <arnaud.porterie@docker.com>

### DIFF
--- a/faq/index.html
+++ b/faq/index.html
@@ -127,14 +127,6 @@ headline: 'Frequently Asked Questions'
                     </li>
 
                     <li class="submenu">
-                        <h6 class="arrow-r">Can I use it in the browser?</h6>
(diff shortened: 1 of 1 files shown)


icecrime:
CLA signed (it's a different email address that I've used in the past indeed).

zinuga:
Thanks for noticing and fixing. LGTM.

PR #366, opened 2016-08-25
Fix formatting of Network Failures table

zandaleph:
Reading http://www.grpc.io/docs/guides/error.html it was clear that this segment had been mangled by an edit.  Reverting it to state before change: https://github.com/grpc/grpc.github.io/commit/076c9c1e351bd099394e7cf557451682f72896df#diff-43cc74a26b167e70dcecef7f8857ad94L38

### DIFF
--- a/docs/guides/error.md
+++ b/docs/guides/error.md
@@ -43,11 +43,8 @@ Server threw an exception (or did something other than returning a status code t
 
 Case | Status code
 -----|-----------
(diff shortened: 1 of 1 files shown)


zandaleph:
I signed it!

LisaFC:
Duplicates https://github.com/grpc/grpc.github.io/pull/364, which I just merged in - but thanks for spotting and fixing!

PR #381, opened 2016-09-09
remove nuget update -self suggestion after mono update

apolcyn:
So indeed it looks like the latest releases of mono and Xamarin Studio update nuget to 2.12.
I guess it's noted in the 4.4.2 and 6.0 release notes for each btw.

For the actual issue with Grpc.Tools and nuget 2.12, filed https://github.com/NuGet/Home/issues/3427#issuecomment-246019865. To work around it, the workaround in this change can work without using nuget, though it might be more involved than desired...

### DIFF
--- a/docs/quickstart/csharp.md
+++ b/docs/quickstart/csharp.md
@@ -35,7 +35,7 @@ On Windows, using Visual Studio, you'll need:
 
 On OS X, using Xamarin Studio, you'll need:
 
-* Mono 4.4.2+ (or Mono 4+ is sufficient if you manually update nuget to version 2.12+) 
+* Mono 4.4.2+ (or Mono 4+ is sufficient if you manually update NuGet to version 2.12+) 
 * Xamarin Studio 6.0+ 
 * Git (to download the sample code)
 
@@ -97,8 +97,10 @@ a NuGet dependency on C# gRPC.
 The problem is that C# gRPC package currently depends on 
 System.Interactive.Async 3.0.0, which requires NuGet 2.12+ to install.
 The NuGet included on the latest versions of Monodevelop is too old to install gRPC C#.
-You can work around this issue by restoring the C# gRPC NuGet dependency from the command line
-, and then proceeding to build and edit from the Monodevelop IDE, as follows:
+
+If you don't want to change the version of NuGet that you're using, 
+a possible workaround to get these files is to download the NuGet 
+package and unzip without a NuGet client, as follows.
 
 * Install NuGet 2.12+ so that it's available from the command line.
 * From the `examples/csharp/helloworld` directory, run `/path/to/nuget restore`. 
@@ -204,7 +206,7 @@ message HelloReply {
 
 Next we need to update the gRPC code used by our application to use the new service definition. 
 
-The Grpc.Tools NuGet package contains the protoc and protobuf C# plugin binaries you will need to generate the code. 
+The `Grpc.Tools` NuGet package contains the protoc and protobuf C# plugin binaries you will need to generate the code. 
 
 ### Obtaining the Grpc.Tools NuGet package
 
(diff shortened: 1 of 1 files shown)


docs/quickstart/csharp.md
```diff
@@ -85,16 +89,16 @@ From the `examples/csharp/helloworld-from-cli` directory:
 > dotnet build **/project.json
 ```
 
-### Using Xamarin Studio or Monodevelop IDEs
-The C# gRPC package currently depends on System.Interactive.Async 3.0.0, which cannot be downloaded on older NuGet versions.
-NuGet is too old in Xamarin Studio on OS X and Monodevelop on Linux as well as via apt-get.
+### Using the Monodevelop IDE
+The C# gRPC package currently depends on System.Interactive.Async 3.0.0, which requires NuGet 2.12+ to install.
+The NuGet included on the latest versions of Monodevelop is too old to install gRPC C#.
```
jtattermusch: what does it mean? that with mono and nuget 2.12 I cannot do the initial restore directly from the monodevelop IDE, but after running 'nuget restore' from the commandline, everything will work as expected when I work with monodevelop afterwards? (if so, things are totally fine I'd say).
We should clarify this.

I'll be fine with running 'nuget restore' from the commandline, but I cannot hit the "restore" button inde tfrom IDE?

docs/quickstart/csharp.md
```diff
@@ -25,26 +25,25 @@ Using the .NET Core SDK on Windows, OS X, or Linux, you'll need:
...
+* .NET Framework 4.5+
+* Visual Studio 2013 or 2015.
 * Git (to download the sample code)
 
 On OS X, using Xamarin Studio, you'll need:
 
-* Mono 4+
-* Xamarin Studio 5.9+
-* A NuGet executable (to download the Grpc.Tools package that contains protoc and protobuf binaries for code generation)
+* Mono 4.4.2+
```
jtattermusch: let's add "(or Mono 4+ is sufficient if you manually update nuget to version 2.12)"

jtattermusch:
Looks mostly good after addressing a few nits.
After that, please reassign to @LisaFC  who will merge once happy.

docs/quickstart/csharp.md
apolcyn: yes this is what I meant, tried to re-word to make this more clear in latest commit

apolcyn:
cc @hsaliak

docs/quickstart/csharp.md
```diff
@@ -199,23 +208,30 @@ The Grpc.Tools NuGet package contains the protoc and protobuf C# plugin binaries
...
-> /path/to/nuget install Grpc.Tools
+$ mkdir packages && cd packages
+$ /path/to/nuget install Grpc.Tools
 ```
 
-* Note that you may have to update your NuGet executable to the latest version
-  in order to install the Grpc.Tools package, depending on your operating system.
+#### If you have a NuGet client that is at version 2.12
+
+NuGet 2.12 does not install the files from the Grpc.Tools package necessary on Linux and OS X.
```
LisaFC: does not install the necessary files from the `Grpc.Tools` package on Linux and OS X.

docs/quickstart/csharp.md
```diff
@@ -199,23 +208,30 @@ The Grpc.Tools NuGet package contains the protoc and protobuf C# plugin binaries
...
+Without changing the version of NuGet that you're using, a possible workaround to obtaining the binaries included in the Grpc.Tools package 
+is by simply downloading the NuGet package and unzipping without a NuGet client, as follows.
```
LisaFC: If you don't want to change the version of NuGet that you're using, a possible workaround to get these files is to download the NuGet package and unzip without a NuGet client, as follows.

docs/quickstart/csharp.md
```diff
@@ -199,23 +208,30 @@ The Grpc.Tools NuGet package contains the protoc and protobuf C# plugin binaries
...
+From the directory of you're example:
```
LisaFC: From your example directory:

docs/quickstart/csharp.md
```diff
@@ -25,26 +25,25 @@ Using the .NET Core SDK on Windows, OS X, or Linux, you'll need:
...
-* Xamarin Studio 5.9+
-* A NuGet executable (to download the Grpc.Tools package that contains protoc and protobuf binaries for code generation)
+* Mono 4.4.2+ (or Mono 4+ is sufficient if you manually update nuget to version 2.12+) 
+* Xamarin Studio 6.0+ 
 * Git (to download the sample code)
 
 On Linux, using the Monodevelop IDE, you'll need:
 
-* Mono 4+
+* Mono 4.4.2+ (or Mono 4+ is sufficient if you manually update nuget to version 2.12+)
```
LisaFC: nuget -> NuGet

PR #384, opened 2016-09-13
Accessibility changes to styling

LisaFC:
Darker text colours, bigger line spacing.

Fixes (hopefully?) https://github.com/grpc/grpc.github.io/issues/354

Google folks on our network can see a preview at http://lcarey0-l.dub.corp.google.com:4000/grpc.github.io/docs/, PTAL.

### DIFF
--- a/_sass/base/_colors.scss
+++ b/_sass/base/_colors.scss
@@ -24,7 +24,7 @@ $linkColorTertiary: $gray;
 $codeColor: darken($linkColor, 20%);
 $codeBkColor: #ebfaf8;
 $wellColor: #F9F9F9;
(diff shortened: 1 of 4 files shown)


hsaliak:
I like the slightly darker version. 
However, one other thing I would like to fix is that the links are weighted(?) slightly higher than normal text, leading to a very distracting read. Eg: http://lcarey0-l.dub.corp.google.com:4000/grpc.github.io/blog/

Any chance we can fix that too in this PR?

LisaFC:
OK, reduced the weight for link text slightly, PTAL.

hsaliak:
Much better, thank you! LGTM

PR #393, opened 2016-10-03
Fix example for ruby server

teodor-pripoae:
Hi,

This fixes the confusing example. It didn't specify what kind `port` has. My first though was that is was an int:

``` ruby
s = GRPC::RpcServer.new
s.add_http2_port(8765, :this_port_is_insecure)

=> TypeError:
       no implicit conversion of Fixnum into String
```

Then, I tried as string:

``` ruby
s = GRPC::RpcServer.new
s.add_http2_port("8765", :this_port_is_insecure)
=> RuntimeError:
       could not add port 8765 to server, not sure why
```

Finally it worked with address string

``` ruby
s = GRPC::RpcServer.new
s.add_http2_port(":8765", :this_port_is_insecure)
```

### DIFF
--- a/docs/tutorials/basic/ruby.md
+++ b/docs/tutorials/basic/ruby.md
@@ -282,9 +282,10 @@ so that clients can actually use our service. The following snippet shows how we
 do this for our `RouteGuide` service:
 
 ```ruby
+addr = "0.0.0.0:8080"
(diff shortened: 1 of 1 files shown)


teodor-pripoae:
Fixed that. Is it ok now ?

PR #394, opened 2016-10-03
Use defined variable in C++ auth example

codegoalie:
The comment does say "using the credentials created in the previous step," but I think it's still better using the same variable name.


codegoalie:
I signed it!

PR #404, opened 2016-10-12
Document that path must be of the same format, regardless of IDL.

markdroth:
As we discussed a couple of weeks ago.  Needed for grpc/grpc#7818.


a11r:
LGTM

Issue #406, opened 2016-10-13
"Samples" link is highlighted incorrectly


tadeegan:
When I click the references tab I also see the "Samples" link seems to be selected.

hsaliak:
Should be addressed by #426 as well.. so closing. Thanks for the PR and issue!

PR #407, opened 2016-10-13
Update selected header

tadeegan:
#406


hsaliak:
#426 solves the same problem, closing this one out since I merged that.

PR #416, opened 2016-11-04
Document thread-safety of a service

jagij:
A service should be implemented in a thread-safe way. A comment on the thread safety of the example service has been added.

### DIFF
--- a/docs/tutorials/basic/c.md
+++ b/docs/tutorials/basic/c.md
@@ -253,8 +253,8 @@ finished dealing with the RPC and that the `Feature` can be returned to the
 client.
 
 Note that all service methods can (and will!) be called from multiple threads at
-the same time. So you have to make sure that your method implementations are
-thread safe. In our example case here, `feature_list_` is never changed after
+the same time. You have to make sure that your method implementations are
+thread safe. In our example, `feature_list_` is never changed after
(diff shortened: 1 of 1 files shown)


jagij:
I signed it!

docs/tutorials/basic/c.md
```diff
@@ -252,6 +252,12 @@ information, and then `return` with an `OK` status to tell gRPC that we've
 finished dealing with the RPC and that the `Feature` can be returned to the
 client.
 
+Note that all service methods can (and will!) be called from multiple threads at
+the same time. So you have to make sure that your method implementations are
```
dgquintas: Remove "So".

docs/tutorials/basic/c.md
```diff
@@ -252,6 +252,12 @@ information, and then `return` with an `OK` status to tell gRPC that we've
...
+thread safe. In our example case here, `feature_list_` is never changed after
```
dgquintas: remove "case here"

jagij:
Thanks for the feedback! I made changes accordingly.
It seems I used a different e-mail address last time, so I amended the commit with the already accepted CLAs, and force pushed it. Hope that doesn't break things..

PR #417, opened 2016-11-07
Fix proto filename in Python quickstart

fowlslegs:
Puts the correct proto filename in place for the Python quickstart example.


fowlslegs:
I signed it!

hsaliak:
This has been addressed by #423, so closing. Thanks for the pull request!

PR #419, opened 2016-11-14
Replace route_guide.proto with helloworld.proto in Python quickstart …

derakon:
…guide.

The current Python quickstart guide has an incorrect command for re-generating the server/client after changing helloworld.proto. I checked the Java, PHP, and Ruby quickstart guides and they don't have this problem; it's just the Python one.

I didn't actually test this, but it's just a text change; what could possibly go wrong? :)


derakon:
Decided to check your other pull requests, and you already have one pending that does this. Please disregard this PR.

PR #422, opened 2016-11-18
Update python.md

chadgates:
example states wrong proto for recreation (route_guide instead of helloworld)


chadgates:
I signed it!

PR #434, opened 2016-12-20
Correct branch name from the repository

jmcshane:
This name does not match the branch name in the repository

### DIFF
--- a/_data/config.yml
+++ b/_data/config.yml
@@ -1,3 +1,3 @@
 grpc_release_branch: v1.0.0
-grpc_java_release_tag: v1.0.0
+grpc_java_release_tag: v1.0.x
 milestones_link: https://github.com/grpc/grpc/milestones


jmcshane:
I signed it!

ejona86:
This is not supposed to be a branch name. It is supposed to be a tag. It's named "grpc_java_release_tag" after all.

However, this does demonstrate that updating the tag name is missing from our [release process](https://github.com/grpc/grpc-java/blob/master/RELEASING.md), so we do need to update the variable here and I'll need to add some documentation to remind us in the future.

ejona86:
I've created #439 to bump the version to the current version, but the variable will still be a tag.

ejona86:
I've created grpc/grpc-java#2567 to avoid forgetting in the future.

ejona86:
@jmcshane, things should work using the tag. Although I do think git spews more messages than if it were a tag. It is important to be a tag though, otherwise the artifacts probably would fail to download (since they'll be SNAPSHOTs instead of the released versions).

However, if something is actually failing as it is now, then please say so.

PR #435, opened 2016-12-21
force list-nested p tags to have inline display

seryl:
Fixes some display issues with lists on the [/about](http://www.grpc.io/about/) page.

## Before

before

## After

after


seryl:
I signed it!

PR #454, opened 2017-02-07
PHP: Update quickstart guide

### LINKED ISSUE #431: Php quickstart ubuntu doesn't work
In the php quickstart, the ubuntu instructions don't work for recent versions of ubuntu http://www.grpc.io/docs/quickstart/php.html#debianubuntu.

php5 is no longer a system installable package by default. Could there be an ubuntu version added to the instructions or an alternate php5 source ?

### LINKED ISSUE #436: PHP: add section to warn first time users on OSX El Capitan
For first time users on OSX El Capitan, users might need to perform a few extra steps before they can install pecl, the PHP extensions repository that can install our gRPC PHP extension. We should call this our in our doc.

Relevant blog post: http://blog.g-design.net/post/137712472685/configuring-apache-and-php-after-updating-to-os-x

### LINKED ISSUE #437: PHP: call out the need to add "extension=grpc.so" to php.ini in our docs
After the user installed our gRPC PHP Extension via the "pecl install grpc" command, the user still needs to add a line "extension=grpc.so" to their respective php.ini config file. We should call this out in our doc.

stanley-cheung:
Fixes #436 
Fixes #437 
Fixes #431

### DIFF
--- a/docs/quickstart/php.md
+++ b/docs/quickstart/php.md
@@ -18,16 +18,21 @@ working example.</p>
 * `pecl`: version 1.9 or higher
 * `composer`
 
-### Debian/Ubuntu
+### Install PHP and PECL on Debian/Ubuntu
+
+PHP 5.5 or above
+
 ```sh
 $ [sudo] apt-get install php5 php5-dev php-pear zlib1g-dev
-
-OR
-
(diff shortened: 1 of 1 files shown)


stanley-cheung:
@LisaFC  Please review, thanks!

PR #465, opened 2017-02-27
fix broken link to multi lang benchmark dashboard

apolcyn:


jtattermusch:
Normally, @LisaFC reviews content changes to the grpc.io site. This is trivial enough, so I'll give and LGTM as well, but you should still wait for Lisa.

jtattermusch:
I'll just merge as the old link is clearly broken.

PR #473, opened 2017-03-15
Fix typo (missing comma) in C++ Quick Start guide

eduherminio:


eduherminio:
Can you get this checked, @hsaliak? It shouldn't take more than a few seconds.

PR #475, opened 2017-03-19
Bump the grpc release version to v1.2.0

mehrdada:
Our guides often ask the user to issue a git
clone command which checkouts a branch specified
in grpc.io/release URL.  It is critical that the
release version gets updated as we ship new
versions. This patch bumps it from 1.0.0 to
1.2.0.

(c.f. https://github.com/grpc/grpc.github.io/issues/471)

### DIFF
--- a/_data/config.yml
+++ b/_data/config.yml
@@ -1,3 +1,3 @@
-grpc_release_branch: v1.0.0
-grpc_java_release_tag: v1.0.3
+grpc_release_branch: v1.2.0
+grpc_java_release_tag: v1.2.0
 milestones_link: https://github.com/grpc/grpc/milestones


_data/config.yml
```diff
@@ -1,3 +1,3 @@
-grpc_release_branch: v1.0.0
+grpc_release_branch: v1.2.0
 grpc_java_release_tag: v1.0.3
```
jtattermusch: java should be also bumped?

_data/config.yml
```diff
@@ -1,3 +1,3 @@
-grpc_release_branch: v1.0.0
```
jtattermusch: btw, v1.0.0 is actually a tag, not a branch, but I guess it's just the naming that wrong.

jtattermusch:
Added @stanley-cheung  and removed Craig and myself.

_data/config.yml
mehrdada: I am reluctant to changing it in this tiny commit as it is referenced in a dozen places and might need more thought and knowledge of project history.

_data/config.yml
mehrdada: Done.

mehrdada:
@stanley-cheung friendly ping.

_data/config.yml
stanley-cheung: So this value of this variable goes into link like this:

"https://github.com/grpc/grpc/blob/v1.2.0/examples/protos/route_guide.proto"

So as long as it can go after the `/blob/` part of the github URL that works. And tags work in that part of the URL. So yes the variable naming might be a bit off (both branch name and tag name can go after `/blob/`) but we can continue with this update.

PR #476, opened 2017-03-20
Move tracking snippet to top of page and load async

carl-mastrangelo:
This has two main benefits:

1.  The tracking snippet needs to be in <head> in order to use
    Google Search Console, which is the only way to see what
    search terms ended up directing users to grpc.io.  Without it,
    the search query shows up as the infamous "(not provided)".

2.  Page loads are faster.  The [analytics][1] page suggests
    putting it in the <head>.  Also, it suggests page loads are
    faster if using a modern browser.  Checkout our Google
    Analytics for grpc.io shows an overwhelming number of users
    with Chrome, Firefox, Safari.  For IE users (a small percent),
    most are using version 11 which supports async js.

[1]: https://developers.google.com/analytics/devguides/collection/analyticsjs/

cc: @LisaFC

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -8,6 +8,13 @@
     <meta name="description" content="A high-performance, open-source universal RPC framework. Client applications can directly call methods on a server application on a different machine as if it was a local object.">
     <title>grpc / {{ page.title }}</title>
 
+    <script>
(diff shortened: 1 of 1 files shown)


jtattermusch:
LGTM.

PR #477, opened 2017-03-21
Small fixes for ruby docs

sethherr:
I'm pretty sure you don't mean python. Also, I'm pretty sure you mean to link to quickstart - not installation, which isn't a thing.


sethherr:
I signed it!

PR #482, opened 2017-03-23
Correct deadlines/timeouts text

nathanielmanistaatgoogle:
In particular it is no longer the case that a timeout must be specified when invoking an RPC with gRPC Python.


nathanielmanistaatgoogle:
Thanks!

PR #484, opened 2017-03-26
Add troubleshooting info for pod install error

avinassh:


avinassh:
This had happened earlier too: #7396

PR #487, opened 2017-04-09
Reposting the grpc & grpc-gateway post from https://medium.com/yik-yak-eng

stxmendez:
Reposting the https://medium.com/yik-yak-eng/migration-to-google-cloud-platform-grpc-grpc-gateway-1d6e626b71a9 blog post here.

Concerns with this PR:
- It appears most posts include a date in their file name so I just picked one.
- I think that the CPU graphs are too big.
- Never used Jekyll, expect noob mistakes.


zinuga:
@hsaliak can you review and merge?

PR #497, opened 2017-04-24
Changes type reference in description

dgruntz:
Changes return type HelloResponse to HelloReply in the description of the helloworld.proto definition.

The description states:

> For now all you need to know is that both the server and the client "stub" have a `SayHello` RPC
> method that takes a `HelloRequest` parameter from the client and returns a `HelloResponse` 
> from the server, and that this method is defined like this:
> 
> // The greeting service definition.
> service Greeter {
>     // Sends a greeting
>     rpc SayHello (HelloRequest) returns (HelloReply) {}
> }

so I think the SayHello RPC method returns a HelloReply from the server.


dgruntz:
Btw, I signed the Google Contributor License Agreements.

PR #501, opened 2017-04-28
Update ruby.md

Jwan622:
Broken tutorial for route_guide server

I left a note in the code. Basically. this command will fail because there is an ARGV check in the actual server file.

### DIFF
--- a/docs/tutorials/basic/ruby.md
+++ b/docs/tutorials/basic/ruby.md
@@ -398,6 +398,13 @@ Run the server, which will listen on port 50051:
 ```
 $ # from examples/ruby
 $ bundle exec route_guide/route_guide_server.rb
+# I don't know what the solution is but this will fail because in the server.rb file, we have this gaurd:
+```
+if ARGV.length == 0
+    fail 'Please specify the path to the route_guide json database'
+  end
+```
+# So what should the ARGV be?
 ```
 Run the client (in a different terminal):
 


LisaFC:
Michael, can you help with this or find someone who can? I don't know enough about Ruby....

murgatroid99:
@apolcyn Was something done recently to improve this?

apolcyn:
actually this tutorial use to refer to the python json file, but it looks like PR: https://github.com/grpc/grpc.github.io/pull/477/files took that out.

After comments in this PR, can we please add back in the json file reference that was recently reverted?

I think that would be good then, thanks!

docs/tutorials/basic/ruby.md
```diff
@@ -398,6 +398,13 @@ Run the server, which will listen on port 50051:
 ```
 $ # from examples/ruby
 $ bundle exec route_guide/route_guide_server.rb
+# I don't know what the solution is but this will fail because in the server.rb file, we have this gaurd:
```
apolcyn: can we please remove this line

docs/tutorials/basic/ruby.md
```diff
@@ -398,6 +398,13 @@ Run the server, which will listen on port 50051:
...
+```
+if ARGV.length == 0
+    fail 'Please specify the path to the route_guide json database'
+  end
+```
+# So what should the ARGV be?
```
apolcyn: I'd like to remove this one too

apolcyn:
Thanks for pointing out the error here. This fix here was just done in https://github.com/grpc/grpc.github.io/pull/531 though, so going to close this.

PR #502, opened 2017-05-02
Doc on Go generated code

apolcyn:
This is a rough draft right now and I'm planning to do some more polishing.

cc @hsaliak @dfawley @ericgribkoff  if the general direction of this looks ok.

### DIFF
--- a/docs/generated/go.md
+++ b/docs/generated/go.md
@@ -1,0 +1,201 @@
+---
+bodyclass: docs
+headline: Go Generated Code Guide
+layout: docs
+sidenav: doc-side-quickstart-nav.html
+type: markdown
+---
+
+<p class="lead">This guide gives an overview of code generated with the protoc grpc go plugin.</p>
+
+<div id="toc"></div>
+
+## Compiler Invocation and Results of Code Generation
+This section provides instructions for using the `grpc` plugin to `protoc-gen-go`, when compiling `.proto` files with `protoc`.
+See https://developers.google.com/protocol-buffers/docs/reference/go-generated for detailed instructions on 
+how to use the `protoc` compiler the `protoc-gen-go` plugin.
+
+When `protoc` is invoked with a `--go_out=plugins=grpc:` argument, gRPC stub and service interface code gets generated
+from the `.proto` files specified in the rest of the command. The names of the output files are computed in the same way as though the "plain" Go
+`protoc-gen-go` was used (as per the description in https://developers.google.com/protocol-buffers/docs/reference/go-generated). But additional content
+is added to the generated `.pb.go` files as follows:
+
+Each `service Foo` in the `.proto` file results in a `FooClient interface` and a `FooServer interface` in the `.pb.go` file.
+  * For example, a declaration of `service Foo` in `bar.proto` results in a `type FooServer interface` and a `type FooClient interface` in `bar.pb.go`.
+  
+Each `rpc` declaration within a `service` definition in the `.proto` files results in a corresponding method on the `Server` and `Client` interfaces.
+Method arguments and return values differ depending on whether the method is on the client or server interface, and on the type of RPC 
+(e.g. unary, client-streaming, server-streaming, bidi-streaming).
+
+### Different RPC types and their RPC signatures in proto files
+The different types RPC's (unary, client-streaming, server-streaming, and bidi-streaming),
+are declared in the `service` declarations in `.proto` files as follows:
+
+Note that type message types `MsgA` and `MsgB` are just placeholders for arbitrary protobuf message types. Also note that the descriptions
+further on this document refer to the example RPC signatures in this subsection.
+
+#### Unary Method signature
+`rpc Foo(MsgA) returns (MsgB)`
+
+#### Server-streaming Method signature
+`rpc Foo(stream MsgA)`
+
+#### Client-streaming Method signature
+`rpc Foo(MsgA) returns (stream MsgB)`
+
+#### Bidi-streaming Method signatures
+`rpc Foo(stream MsgA) returns (stream MsgB)`
+    
+## Using the generated service interfaces:
+
+#### Implementation and Registration of a service definition
+On server side, each `service Bar` in the `.proto` file results in the function: `func RegisterFooServer(s *grpc.Server, srv BarServer)`.
+The application can define a concrete implementation of the `BarServer` interface and register it with a `grpc.Server` instance 
+(before starting the server instance) by using this function.
+
+#### Unary methods
+These methods have the following signature on the generated service interface:
+`Foo(context.Context, *MsgA) (*MsgB, error)`
+
+In this context, `MsgA` is the protobuf message sent from the client, and `MsgB` is the protobuf message sent back from the server.
+
+#### Server-streaming methods
+These methods have the following signature on the generated service interface:
+`Foo(*MsgA, <ServiceName>_FooServer) error`
+
+In this context, `MsgA` is the single request from the client, and the `<ServiceName>_FooServer` parameter represents the server-to-client stream
+of `MsgB` messages.
+
+`<ServiceName>_FooServer` has an embedded `grpc.ServerStream` and the following interface:
+
+```
+type <ServiceName>_FooServer interface {
+	Send(*MsgB) error
+	grpc.ServerStream
+}
+```
+
+The server-side handler can send a stream of protobuf messages to the client through this parameters' `Send` method. End-of-stream for the server-to-client
+stream is caused by the `return` of the handler method.
+  
+#### Client-streaming methods
+These methods have the following signature on the generated service interface:
+`Foo(<ServiceName>_FooServer) error`
+
+In this context, `<ServiceName>_FooServer` can be used both to read the client-to-server message stream and to send the single server response message.
+
+`<ServiceName>_FooServer` has an embedded `grpc.ServerStream` and the following interface:
+
+```
+type <ServiceName>_FooServer interface {
+	SendAndClose(*MsgA) error
+	Recv() (*MsgB, error)
+	grpc.ServerStream
+}
+```
+
+The server-side handler can repeatedly call `Recv` on this parameter in order to receive the full stream of
+messages from the client. `Recv` returns `(nil, io.EOF)` once it has reached the end of the stream.
+The single response message from the server is sent by calling the `SendAndClose` method on this `<ServiceName>_FooServer` parameter. 
+Note that `SendAndClose` must be called once and only once.
+  
+#### Bidi-streaming methods
+These methods have the following signature on the generated service interface:
+`Foo(<ServiceName>_FooServer) error`
+
+In this context, `<ServiceName>_FooServer` can be used to access both the client-to-server message stream and the server-to-client message stream.
+`<ServiceName>_FooServer` has an embedded `grpc.ServerStream` and the following interface:
+
+```
+type <ServiceName>_FooServer interface {
+	Send(*MsgA) error
+	Recv() (*MsgB, error)
+	grpc.ServerStream
+}
+```
+
+The server-side handler can repeatedly call `Recv` on this parameter in order to read the client-to-server message stream.
+`Recv` returns `(nil, io.EOF)` once it has reached the end of the client-to-server stream.
+The response server-to-client message stream is sent by repeatedly calling the `Send` method of on this `ServiceName>_FooServer` parameter.
+End-of-stream for the server-to-client stream is indicated by the `return` of the bidi method handler.
+
+### Methods on generated client interfaces:
+For client side usage, each `service Bar` in the `.proto` file also results in the function: `func BarClient(cc *grpc.ClientConn) BarClient`, which
+returns a concrete implementation of the `BarClient` interface (this concrete implementation also lives in the generated `.pb.go` file).
+
+#### Unary Methods 
+These methods have the following signature on the generated client stub:
+`(ctx context.Context, in *MsgA, opts ...grpc.CallOption) (*MsgB, error)`.
+
+In this context, `MsgA` is the single request from client to server, and `MsgB` contains the response sent back from the server.
+
+#### Server-Streaming methods
+These methods have the following signature on the generated client stub:
+`Foo(ctx context.Context, in *MsgA, opts ...grpc.CallOption) (<ServiceName>_FooClient, error)`.
+
+In this context, `<ServiceName>_FooClient` represents the server-to-client `stream` of `MsgB` messages.
+
+This stream has an embedded `grpc.ClientStream` and the following interface:
+
+```
+type <ServiceName>_FooClient interface {
+	Recv() (*MsgB, error)
+	grpc.ClientStream
+}
+```
+
+The stream begins when the client calls the `Foo` method on the stub. 
+The client can then repeatedly call the `Recv` method on the returned `<ServiceName>_FooClient` return value in order to read the server-to-client response stream. 
+This `Recv` method returns `(nil, io.EOF)` once the server-to-client stream has been completely read through.
+
+#### Client-Streaming methods
+These methods have the following signature on the generated client stub:
+`Foo(ctx context.Context, opts ...grpc.CallOption) (<ServiceName>_FooClient, error)`.
+
+In this context, `<ServiceName>_FooClient` represents the client-to-server `stream` of `MsgA` messages.
+
+`<ServiceName>_FooClient` has an embedded `grpc.ClientStream` and the following interface:
+
+```
+type <ServiceName>_FooClient interface {
+	Send(*MsgA) error
+	CloseAndRecv() (*MsgA, error)
+	grpc.ClientStream
+}
+```
+
+The stream begins when the client calls the `Foo` method on the stub. 
+The client can then repeatedly call the `Send` method on the returned `<ServiceName>_FooClient` stream in order to send the client-to-server message stream.
+The `CloseAndRecv` method on this stream must be called once and only once, in order to both close the client-to-server stream
+and receive the single response message from the server.
+
+#### Bidi-Streaming methods:
+These methods have the following signature on the generated client stub:
+`Foo(ctx context.Context, opts ...grpc.CallOption) (<ServiceName>_FooClient, error)`.
+
+In this context, `<ServiceName>_FooClient` represents both the client-to-server and server-to-client message streams.
+
+`<ServiceName>_FooClient` has an embedded `grpc.ClientStream` and the following interface:
+
+```
+type <ServiceName>_FooClient interface {
+	Send(*MsgA) error
+	Recv() (*MsgB, error)
+	grpc.ClientStream
+}
+```
+
+The stream begins when the client calls the `Foo` method on the stub.
+The client can then repeatedly call the `Send` method on the returned `<SericeName>_FooClient` stream in order to send the 
+client-to-server message stream. The client can also repeatedly call `Recv` on this stream in order to
+receive the full server-to-client message stream.
+
+End-of-stream for the server-to-client stream is indicated by a return value of `(nil, io.EOF)` on the `Recv` method of the stream.
+End-of-stream for the client-to-server stream can be indicated from the client by calling the `CloseSend` method on the stream.
+
+### Packages and Namespaces
+When the `prococ` compiler is invoked with `--go_out=plugins=grpc:`, the `proto package` to `Go package` "translation"
+works the same as when the `protoc-gen-go` plugin is used without the `grpc` plugin (as described in https://developers.google.com/protocol-buffers/docs/reference/go-generated).
+
+So, for example, if `foo.proto` declares itself to be in `package foo`, then the generated `foo.pb.go` file will also be in
+the "Go package" `foo`.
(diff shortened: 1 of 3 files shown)


mehrdada:
@apolcyn Would you mind moving the `go.md` file to `generatedcode` subdirectory for consistency? That's where my merged python guide ended up landing: http://www.grpc.io/docs/generatedcode/python.html

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
+---
+bodyclass: docs
+headline: Go Generated Code Guide
+layout: docs
+sidenav: doc-side-quickstart-nav.html
```
mehrdada: Quickstart sidenav is probably the one we want for reference docs. (I ended up removing the nav line for python for the time being.)

LisaFC: You actually probably don't want the quickstart sidenav as otherwise your reference doc will end up with the quickstarts. :)

What we probably want is a nice new doc-side-reference-nav.html - at the moment we don't have one because we only have a single page in the reference section linking to the autogenerated docs, but if we're going to start to have more manually added reference pages we'll need one for users to navigate through the section.

Let me have a think about how to incorporate the existing reference landing page into a structure with a side nav and I'll get back to you...

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+type: markdown
+---
+
+<p class="lead">This guide gives an overview of code generated with the protoc grpc go plugin.</p>
+
+<div id="toc"></div>
+
+## Compiler Invocation and Results of Code Generation
+This section provides instructions for using the `grpc` plugin to `protoc-gen-go`, when compiling `.proto` files with `protoc`.
```
LisaFC: This section provides instructions for using the `grpc` plugin when generating code from .proto files with `protoc` and `protoc-gen-go`.

(just generally, this sounds like grpc for Go is a plugin to protoc-gen-go, which is a plugin for protoc - can a plugin have a plugin?)

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+See https://developers.google.com/protocol-buffers/docs/reference/go-generated for detailed instructions on 
+how to use the `protoc` compiler the `protoc-gen-go` plugin.
```
LisaFC: See the Protocol Buffers [Go Generated Code Guide](https://developers.google.com/protocol-buffers/docs/reference/go-generated) for detailed instructions on 
how to use `protoc` compiler with the `protoc-gen-go` plugin.

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+
+When `protoc` is invoked with a `--go_out=plugins=grpc:` argument, gRPC stub and service interface code gets generated
```
LisaFC: gets generated -> is generated

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+from the `.proto` files specified in the rest of the command. The names of the output files are computed in the same way as though the "plain" Go
```
LisaFC: The names of the output files are computed in the same way as when generating the default protocol buffers code with `protoc-gen-go`.

(we don't need to send them to the protobuf Generated Code guide again, we've already told them where it is)

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+`protoc-gen-go` was used (as per the description in https://developers.google.com/protocol-buffers/docs/reference/go-generated). But additional content
+is added to the generated `.pb.go` files as follows:
```
LisaFC: But additional content -> However, additional content

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+
+Each `service Foo` in the `.proto` file results in a `FooClient interface` and a `FooServer interface` in the `.pb.go` file.
```
LisaFC: I'd make this into a bulleted list:

* Each `service Foo` in the `.proto` file results in a `FooClient interface` and a `FooServer interface` in the `.pb.go` file. For example, a declaration of `service Foo` in `bar.proto` results in a `type FooServer interface` and a `type FooClient interface` in `bar.pb.go`.

* Each `rpc` declaration within a `service` definition in the `.proto` files results in a corresponding method on the `Server` and `Client` interfaces. Method arguments and return values differ depending on whether the method is on the client or server interface, and on the type of RPC: unary, client-streaming, server-streaming, or bidi-streaming.

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+  * For example, a declaration of `service Foo` in `bar.proto` results in a `type FooServer interface` and a `type FooClient interface` in `bar.pb.go`.
+  
+Each `rpc` declaration within a `service` definition in the `.proto` files results in a corresponding method on the `Server` and `Client` interfaces.
+Method arguments and return values differ depending on whether the method is on the client or server interface, and on the type of RPC 
+(e.g. unary, client-streaming, server-streaming, bidi-streaming).
+
+### Different RPC types and their RPC signatures in proto files
```
LisaFC: I don't think we need this section at all. This is basically "how to define a service in a proto file", which isn't Go-specific and is already covered here: http://www.grpc.io/docs/guides/concepts.html#service-definition

docs/generated/go.md
LisaFC: I'd also maybe put this in docs/reference rather than docs/generated, it means we have a section for various reference docs that might come up *including* generated code (eg in the current protocol buffers docs, the reference section includes proto language specs as well as generated code guides and API docs)

docs/generated/go.md
LisaFC: Also add (after removing section below): 

You can find out how to define a gRPC service in a .proto file in [Service Definitions](../guides/concepts.html#service-definition).

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+#### Client-streaming Method signature
+`rpc Foo(MsgA) returns (stream MsgB)`
+
+#### Bidi-streaming Method signatures
+`rpc Foo(stream MsgA) returns (stream MsgB)`
+    
+## Using the generated service interfaces:
+
+#### Implementation and Registration of a service definition
+On server side, each `service Bar` in the `.proto` file results in the function: `func RegisterFooServer(s *grpc.Server, srv BarServer)`.
```
LisaFC: On server side -> On the server side

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+The application can define a concrete implementation of the `BarServer` interface and register it with a `grpc.Server` instance
```
LisaFC: Little code snippet to illustrate this would be nice?

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+`<ServiceName>_FooServer` has an embedded `grpc.ServerStream` and the following interface:
+
+```
+type <ServiceName>_FooServer interface {
+	Send(*MsgB) error
+	grpc.ServerStream
+}
+```
+
+The server-side handler can send a stream of protobuf messages to the client through this parameters' `Send` method. End-of-stream for the server-to-client
```
LisaFC: parameters' -> parameter's

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+	grpc.ServerStream
+}
+```
+
+The server-side handler can repeatedly call `Recv` on this parameter in order to read the client-to-server message stream.
+`Recv` returns `(nil, io.EOF)` once it has reached the end of the client-to-server stream.
+The response server-to-client message stream is sent by repeatedly calling the `Send` method of on this `ServiceName>_FooServer` parameter.
+End-of-stream for the server-to-client stream is indicated by the `return` of the bidi method handler.
+
+### Methods on generated client interfaces:
```
LisaFC: Agree, the structure should be parallel on the server and client sides.

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+The client can then repeatedly call the `Send` method on the returned `<SericeName>_FooClient` stream in order to send the 
+client-to-server message stream. The client can also repeatedly call `Recv` on this stream in order to
+receive the full server-to-client message stream.
+
+End-of-stream for the server-to-client stream is indicated by a return value of `(nil, io.EOF)` on the `Recv` method of the stream.
+End-of-stream for the client-to-server stream can be indicated from the client by calling the `CloseSend` method on the stream.
+
+### Packages and Namespaces
+When the `prococ` compiler is invoked with `--go_out=plugins=grpc:`, the `proto package` to `Go package` "translation"
+works the same as when the `protoc-gen-go` plugin is used without the `grpc` plugin (as described in https://developers.google.com/protocol-buffers/docs/reference/go-generated).
```
LisaFC: the `proto package` to Go package translation works in the same way as when using `protoc-gen-go` without the `grpc` plugin.

(you can take out the "as described in..." bit)

docs/generated/go.md
```diff
@@ -0,0 +1,201 @@
...
+
+So, for example, if `foo.proto` declares itself to be in `package foo`, then the generated `foo.pb.go` file will also be in
+the "Go package" `foo`.
```
LisaFC: the Go package `foo`.

docs/generated/go.md
mehrdada: > You actually probably don't want the quickstart sidenav as otherwise your reference doc will end up with the quickstarts. :)

Right. That's what I meant, but forgot "don't". Sorry :)

docs/generated/go.md
apolcyn: Ok, I ended up just deleted the code generation command instructions below `Compiler Invocation and Results of Code Generation`

docs/generated/go.md
apolcyn: for below comment, I'm moving this to `docs/reference`

docs/generated/go.md
hsaliak: Should we discuss goroutine safety in each of these sections? Ie: Are they designed to be called safely from multiple goroutines, if not, what are the considerations.

MakMukhi:
Note on thread safety looks good.

docs/index.html
```diff
@@ -37,6 +37,7 @@ <h2 class="col-sm-8 doc-header">gRPC by Language</h2>
           <a class="lang-card-link" href="quickstart/go.html">Quick Start Guide</a>
           <a class="lang-card-link" href="tutorials/basic/go.html">gRPC Basics Tutorial</a>
           <a class="lang-card-link" href="https://godoc.org/google.golang.org/grpc">API Reference</a>
+          <a class="lang-card-link" href="generated/go.html">Generated Code Guide</a>
```
LisaFC: Need to change link to "reference/go.html"

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+<div id="toc"></div>
+
+## Compiler Invocation and Results of Code Generation
+This section provides instructions for using the [grpc plugin](https://godoc.org/github.com/golang/protobuf/protoc-gen-go/grpc) to `protoc-gen-go`, when compiling `.proto` files with `protoc`.
+See the Protocol Buffers [Go Generated Code Guide](https://developers.google.com/protocol-buffers/docs/reference/go-generated) for detailed instructions on 
+how to use `protoc` compiler with the `protoc-gen-go` plugin.
+
+You can find out how to define a gRPC service in a .proto file in [Service Definitions](../guides/concepts.html#service-definition).
+
+### Different RPC types and their RPC signatures in proto files
```
LisaFC: I thought we were going to remove this entire section, as per my previous review? It just duplicates the information in Service Definitions and isn't Go-specific.

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+#### Server-streaming Method signature
+`rpc Foo(MsgA) returns (stream MsgB)`
+
+#### Client-streaming Method signature
+`rpc Foo(stream MsgA) returns (MsgB)`
+
+#### Bidi-streaming Method signatures
+`rpc Foo(stream MsgA) returns (stream MsgB)`
+    
+## Using the generated service interfaces:
```
LisaFC: Actually, I've just realized almost the entire doc is now inside one giant h2 section (this one) so I think we can just remove this heading.

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Methods on generated server interfaces
```
LisaFC: Change to ##

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+	grpc.ClientStream
+}
+```
+
+The stream begins when the client calls the `Foo` method on the stub. 
+The client can then repeatedly call the `Send` method on the returned `<ServiceName>_FooClient` stream in order to send the client-to-server message stream.
+The `CloseAndRecv` method on this stream must be called once and only once, in order to both close the client-to-server stream
+and receive the single response message from the server.
+
+#### Bidi-Streaming methods:
```
LisaFC: Change to ###, remove :

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+You can find out how to define a gRPC service in a .proto file in [Service Definitions](../guides/concepts.html#service-definition).
+
+### Different RPC types and their RPC signatures in proto files
+The different types RPC's (unary, client-streaming, server-streaming, and bidi-streaming),
+are declared in the `service` declarations in `.proto` files as follows:
+
+Note that type message types `MsgA` and `MsgB` are just placeholders for arbitrary protobuf message types. Also note that the descriptions
+further on this document refer to the example RPC signatures in this subsection.
+
+Thread-safety: note that client-side RPC invocations and server-side RPC handlers <i>are thread-safe</i> and are meant
```
LisaFC: We actually have CSS for a fancy note box (fancy!), so we should do:

```
<p class="note"><strong>Thread-safety</strong>: Client-side RPC invocations and server-side RPC handlers <i>are thread-safe</i> and are meant to be ran on concurrent go routines. However, for <i>individual streams</i>, incoming and outgoing data is bi-directional but serial; so, for example, <i>individual streams</i> do not support <i>concurrent reads</i> or <i>concurrent writes</i> (but reads are safely concurrent <i>with</i> writes).</p>
```

I know we're removing the section this is in but I think it works fine as a note after the main introduction too.

LisaFC: (I had to put the text in a code snippet so you could see the markup, obviously we just want it as HTML in the file!)

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
+---
+bodyclass: docs
+headline: Go Generated Code Guide
+layout: docs
+type: markdown
+---
+
+<p class="lead">This guide gives an overview of code generated with the protoc grpc go plugin.</p>
+
+<div id="toc"></div>
```
LisaFC: you can take this out, as the right hand side TOC div is already in the header.

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Compiler Invocation and Results of Code Generation
```
LisaFC: Don't need this heading.

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+This section provides instructions for using the [grpc plugin](https://godoc.org/github.com/golang/protobuf/protoc-gen-go/grpc) to `protoc-gen-go`, when compiling `.proto` files with `protoc`.
```
LisaFC: This section doesn't really give you instructions for that, it just points you at some other docs.

I'd change it to:

This guide describes the code generated with the [grpc plugin](https://godoc.org/github.com/golang/protobuf/protoc-gen-go/grpc) to `protoc-gen-go` when compiling `.proto` files with `protoc`.

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+`rpc Foo(stream MsgA) returns (stream MsgB)`
+    
+## Using the generated service interfaces:
+
+#### Methods on generated server interfaces
+On the server side, each `service Bar` in the `.proto` file results in the function: `func RegisterBarServer(s *grpc.Server, srv BarServer)`.
+The application can define a concrete implementation of the `BarServer` interface and register it with a `grpc.Server` instance 
+(before starting the server instance) by using this function.
+
+#### Unary methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+These methods have the following signature on the generated service interface:
+`Foo(context.Context, *MsgA) (*MsgB, error)`
+
+In this context, `MsgA` is the protobuf message sent from the client, and `MsgB` is the protobuf message sent back from the server.
+
+#### Server-streaming methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+type <ServiceName>_FooServer interface {
+	Send(*MsgB) error
+	grpc.ServerStream
+}
+```
+
+The server-side handler can send a stream of protobuf messages to the client through this parameter's `Send` method. End-of-stream for the server-to-client
+stream is caused by the `return` of the handler method.
+  
+#### Client-streaming methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+	grpc.ServerStream
+}
+```
+
+The server-side handler can repeatedly call `Recv` on this parameter in order to receive the full stream of
+messages from the client. `Recv` returns `(nil, io.EOF)` once it has reached the end of the stream.
+The single response message from the server is sent by calling the `SendAndClose` method on this `<ServiceName>_FooServer` parameter. 
+Note that `SendAndClose` must be called once and only once.
+  
+#### Bidi-streaming methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+	grpc.ServerStream
+}
+```
+
+The server-side handler can repeatedly call `Recv` on this parameter in order to read the client-to-server message stream.
+`Recv` returns `(nil, io.EOF)` once it has reached the end of the client-to-server stream.
+The response server-to-client message stream is sent by repeatedly calling the `Send` method of on this `ServiceName>_FooServer` parameter.
+End-of-stream for the server-to-client stream is indicated by the `return` of the bidi method handler.
+
+### Methods on generated client interfaces:
```
LisaFC: Change to ##, remove :

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+For client side usage, each `service Bar` in the `.proto` file also results in the function: `func BarClient(cc *grpc.ClientConn) BarClient`, which
+returns a concrete implementation of the `BarClient` interface (this concrete implementation also lives in the generated `.pb.go` file).
+
+#### Unary Methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+For client side usage, each `service Bar` in the `.proto` file also results in the function: `func BarClient(cc *grpc.ClientConn) BarClient`, which
+returns a concrete implementation of the `BarClient` interface (this concrete implementation also lives in the generated `.pb.go` file).
+
+#### Unary Methods 
+These methods have the following signature on the generated client stub:
+`(ctx context.Context, in *MsgA, opts ...grpc.CallOption) (*MsgB, error)`.
+
+In this context, `MsgA` is the single request from client to server, and `MsgB` contains the response sent back from the server.
+
+#### Server-Streaming methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+	Recv() (*MsgB, error)
+	grpc.ClientStream
+}
+```
+
+The stream begins when the client calls the `Foo` method on the stub. 
+The client can then repeatedly call the `Recv` method on the returned `<ServiceName>_FooClient` <i>stream</i> in order to read the server-to-client response stream. 
+This `Recv` method returns `(nil, io.EOF)` once the server-to-client stream has been completely read through.
+
+#### Client-Streaming methods
```
LisaFC: Change to ###

docs/reference/go/go.md
```diff
@@ -0,0 +1,194 @@
...
+
+The stream begins when the client calls the `Foo` method on the stub.
+The client can then repeatedly call the `Send` method on the returned `<SericeName>_FooClient` stream in order to send the 
+client-to-server message stream. The client can also repeatedly call `Recv` on this stream in order to
+receive the full server-to-client message stream.
+
+End-of-stream for the server-to-client stream is indicated by a return value of `(nil, io.EOF)` on the `Recv` method of the stream.
+End-of-stream for the client-to-server stream can be indicated from the client by calling the `CloseSend` method on the stream.
+
+### Packages and Namespaces
```
LisaFC: Change to ##

docs/reference/go/go.md
apolcyn: sorry meant to remove earlier, removed

apolcyn:
tried to respond to all comments again. 

PTAL, I think this is ready again now.

LisaFC:
Thanks for that!

PR #509, opened 2017-05-09
Simplify protoc-gen-go and proto installation

rakyll:


dfawley:
What's the motivation behind this change?  This was lifted directly from the instructions in the https://github.com/golang/protobuf readme:

> The simplest way is to run `go get -u github.com/golang/protobuf/{proto,protoc-gen-go}`. The compiler plugin, protoc-gen-go, will be installed in $GOBIN, defaulting to $GOPATH/bin. It must be in your $PATH for the protocol compiler, protoc, to find it.

Really, `go get -u github.com/golang/protobuf/protoc-gen-go` should be sufficient, because protoc-gen-go imports proto.

rakyll:
Sorry, this was the fault of protobuf's README that is fixed now: https://github.com/golang/protobuf/pull/345.

Curlied go get arguments are the worsts, users are not truly familiar with them. We often see them trying to parse the argument and individually go get the packages rather than copy pasting the instruction. We probably can also do a better job explaining they are valid arguments, but in this case, getting protoc-gen-go is sufficient.

PR #512, opened 2017-05-10
Bump /release version to 1.3.1

mehrdada:


hsaliak:
@dgquintas we should add updating this to the release process. We seem to be missing this and it's necessary for the right docs to be referenced on our site.

PR #519, opened 2017-05-19
Use status.ok() instead of status.IsOk()

drigz:
I needed this change to build. It's also in the docs here: http://www.grpc.io/grpc/cpp/classgrpc_1_1_status.html#a1f5b65c54d4e6dd502897e36040714dc


thisisnotapril:
Closing due to age and no LF CLA

PR #521, opened 2017-05-24
Fix message name in `Greeter` example

soupault:


soupault:
Closed in advance of #522 .

PR #522, opened 2017-05-26
Rename message in `Greeter` examples

soupault:
The message names in the desciprtion and in the code don't match. Seems to be a miss of the previous refactoring.

I've also renamed the messages in `docs/guides/concepts.md` to be consistent with other examples.


soupault:
Any comments on this?

PR #523, opened 2017-05-30
Updating protoc cmd as per example folder structure

javapriyan:
Edited the protoc command to address the example folder structure.


javapriyan:
I signed it!

PR #525, opened 2017-06-07
PHP: Update protobuf runtime library instructions

stanley-cheung:
This is to update the PHP quickstart guide after we de-coupled the protobuf runtime library from our composer package.

FYI @betterbrent


stanley-cheung:
@LisaFC friendly ping. We will need this for the upcoming 1.4 release. Thanks!

bshaffer:
LGTM

PR #527, opened 2017-06-09
Clone examples for wrapped langs from the current release branch.

jtattermusch:
Fixes issue mentioned in https://github.com/grpc/grpc/issues/11426.

### DIFF
--- a/_data/config.yml
+++ b/_data/config.yml
@@ -1,3 +1,3 @@
-grpc_release_branch: v1.3.2
+grpc_release_branch: v1.3.x
 grpc_java_release_tag: v1.3.0
 milestones_link: https://github.com/grpc/grpc/milestones
--- a/docs/quickstart/objective-c.md
+++ b/docs/quickstart/objective-c.md
@@ -55,13 +55,12 @@ Quickstart. Copy the source code from Github
 [repository](https://github.com/grpc/grpc):
 
 ```sh
-$ git clone -b {{ site.data.config.grpc_release_branch }} https://github.com/grpc/grpc.git
+$ git clone --recursive -b {{ site.data.config.grpc_release_branch }} https://github.com/grpc/grpc.git
 ```
 
 ## Install gRPC plugins and libraries
 ```sh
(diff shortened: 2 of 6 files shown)


docs/quickstart/php.md
```diff
@@ -101,7 +101,7 @@ The simplest way to do this is to download pre-compiled binaries for your platfo
 To compile the gRPC PHP Protoc Plugin:
 
 ```sh
-$ git clone https://github.com/grpc/grpc
+$ git clone -b {{ site.data.config.grpc_release_branch }} https://github.com/grpc/grpc
```
mehrdada: How about

```sh
git clone --recursive -b {{ site.data.config.grpc_release_branch }} https://github.com/grpc/grpc
```

instead of a clone and a subsequent submodule init?

docs/quickstart/objective-c.md
```diff
@@ -55,7 +55,7 @@ Quickstart. Copy the source code from Github
 [repository](https://github.com/grpc/grpc):
 
 ```sh
-$ git clone https://github.com/grpc/grpc.git
+$ git clone -b {{ site.data.config.grpc_release_branch }} https://github.com/grpc/grpc.git
```
mehrdada: Similarly, I suggest adding `--recursive` here and removing the line in the following section that clones the submodules.

_data/config.yml
```diff
@@ -1,3 +1,3 @@
-grpc_release_branch: v1.3.2
+grpc_release_branch: v1.3.x
```
jtattermusch: One thing to point out is that the grpc code that's running as part of our examples is not taken for this repository - most examples are using a package from a package manager. The only part the examples can get broken in the release branch is when someone updates the examples themselves.  And the only time we would be updating the examples directly in the release branch is when they're broken and we want to fix them (otherwise we'd just make the change on upstream/master).

So I'd still argue that using 1.3.x has negligible risk of breaking the examples and it saves us some work because we only need to update the branch when there's a new minor release.

PR #530, opened 2017-06-14
Update auth docs for secure ruby connections.

augustoroman:
The existing ruby documentation is out-of-date (uses Credentials instead of ChannelCredentials) and doesn't document the arguments for secure connections.  Unfortunately, there is no auto-generated ruby documentation that I know of that actually pulls together the constructor args and the call options, so once I figured it out I wanted to add it here so that this documentation is useful when creating secure RPCs.

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -189,10 +189,29 @@ stub = Helloworld::Greeter::Stub.new('localhost:50051', :this_channel_is_insecur
 
 #### With server authentication SSL/TLS
 
+See [ChannelCredentials constructor](https://github.com/grpc/grpc/blob/master/src/ruby/ext/grpc/rb_channel_credentials.c#L127..L140) for allowed options.
+
 ```ruby
-creds = GRPC::Core::Credentials.new(load_certs)  # load_certs typically loads a CA roots file
+ca_cert = File.read('ca.pem')
+client_key = File.read('client.key')
+client_cert = File.read('client.pem')
+creds = GRPC::Core::ChannelCredentials.new()
+creds = GRPC::Core::ChannelCredentials.new(ca_cert)
+creds = GRPC::Core::ChannelCredentials.new(ca_cert, client_key, client_cert)
 stub = Helloworld::Greeter::Stub.new('myservice.example.com', creds)
 ```
+
+You can also specify additional channel options using:
+```ruby
+creds = GRPC::Core::ChannelCredentials.new(ca_cert, client_key, client_cert)
+stub = Helloworld::Greeter::Stub.new('1.2.3.4:56789', creds, channel_args: {
+  # override the hostname if you are connecting to a specific IP and using
+  # custom root CA and signed certs on both client and server.
+  GRPC::Core::Channel::SSL_TARGET => "myservice.example.com",
+})
+```
+
+See also the [list of available channel options](https://github.com/grpc/grpc/blob/master/include/grpc/impl/codegen/grpc_types.h#L242..L249) and the [variables exported under GRPC::Core::Channel](https://github.com/grpc/grpc/blob/master/src/ruby/ext/grpc/rb_channel.c#L811..L819).
 
 #### Authenticate with Google
 
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -201,17 +201,24 @@ creds = GRPC::Core::ChannelCredentials.new(ca_cert, client_key, client_cert)
 stub = Helloworld::Greeter::Stub.new('myservice.example.com', creds)
 ```
 
-You can also specify additional channel options using:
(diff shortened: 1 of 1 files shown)


docs/guides/auth.md
```diff
@@ -189,11 +189,30 @@ stub = Helloworld::Greeter::Stub.new('localhost:50051', :this_channel_is_insecur
...
 stub = Helloworld::Greeter::Stub.new('myservice.example.com', creds)
 ```
 
+You can also specify additional channel options using:
+```ruby
+creds = GRPC::Core::ChannelCredentials.new(ca_cert, client_key, client_cert)
+stub = Helloworld::Greeter::Stub.new('1.2.3.4:56789', creds, channel_args: {
+  # override the hostname if you are connecting to a specific IP and using
+  # custom root CA and signed certs on both client and server.
+  GRPC::Core::Channel::SSL_TARGET => "myservice.example.com",
```
apolcyn: per comments in https://github.com/grpc/grpc/pull/10529#pullrequestreview-44169067, I'm not sure about adding in this section as a general guide, can we remove this section?

augustoroman: I'm happy to remove the SSL_TARGET example, but I think adding channel_args examples is important.  The "SSL_TARGET" symbol itself is interesting since it (and only a handful of other channel_arg options) are specifically defined in `GRPC::Core::Channel`.  Perhaps I can do a kitchen-sink example where I dump as many potential options as possible to show what can be done?  (e.g. including `timeout`, which isn't a channel_arg.)

apolcyn: Just in case for clarity, note the channel args constants in ruby still only delegate to defined channel args in https://github.com/grpc/grpc/blob/master/include/grpc/impl/codegen/grpc_types.h (we just have constants for a few likely-frequently-used ones). 

But otherwise more examples of channel args sounds good to me  - but note we wouldn't need to have too much in one example :), as I think there's a lot of different functionality to cover

apolcyn: also I'd still lean towards comments about the SSL_TARGET being intended for tests, rather than examples of it's general usage (same issue as mentioned in #10529)

augustoroman: Updated, let me know if the new example code makes sense.

apolcyn:
@augustoroman I have a couple of hesitations. One is that I don't think the `auth` guide is best for documenting these other timeout-related channel args. Also, while this example does show existence of these channel args, IMO it would be best if an example that used these channel args had an obvious use for them (as it's not necessary to use them in most cases, I want to avoid possibility of making it seem like they are necessary).

augustoroman:
Yeah, I was thinking the same thing about the relevance to the auth guide when I finished making the example, and I agree with your concern that it'll make these args seem necessary when they really are not except in unusual circumstances. I'd like to provide these docs somewhere, but looking through the grpc.io, I didn't see a good existing home for them. I'm happy to (1) trim these examples to only auth-related stuff (including SSL_TARGET but a warning that it's not for prod, but excluding unrelated options such as timeout), (2) trim + make a new page document some of the generated channel semantics somewhere, or ideally (3) update the [autogenerated documentation](http://www.rubydoc.info/gems/grpc/index) if you can give me a hint as to how to get the C code documentation included in the rubydoc, or even manually augmenting the rubydocs? Or maybe (4) something else? What do you recommend?

apolcyn:
```
(1) trim these examples to only auth-related stuff (including SSL_TARGET but a warning that it's not for prod, but excluding unrelated options such as timeout)
```

I have a preference for documenting `SSL_TARGET` along with ruby-specific or "testing" docs, as I'm still uneasy placing it with the general `auth` guide. (this is also since it looks like future work on the SSL stack can make this void for non-testing cases).

```
(2) trim + make a new page document some of the generated channel semantics somewhere, or ideally 
```

Although not obvious, <i>all</i> of the channel args are actually documented with in C/C++ docs (see http://www.grpc.io/grpc/cpp/group__grpc__arg__keys.html - they're defined a couple of headers like https://github.com/grpc/grpc/blob/master/include/grpc/impl/codegen/grpc_types.h), I think that ruby docs should just refer to these when possible. Feel free to update missing docs/references!


```
(3) update the autogenerated documentation if you can give me a hint as to how to get the C code documentation included in the rubydoc, or even manually augmenting the rubydocs?
```

This would be really helpful! I was actually meaning to get to this done sometime soon. I wouldn't suggest spending too much time on it, but if this is doable then feel free to open a PR.

thisisnotapril:
Closing due to age and lack of LF CLA

thelinuxfoundation:

* User @augustoroman isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

PR #538, opened 2017-07-01
Fix generated file name

darkofabijan:
`lib/helloworld_services.rb` => `lib/helloworld_services_pb.rb`


darkofabijan:
I signed it!

PR #539, opened 2017-07-05
Drop www from grpc.io

mehrdada:
Serve the contents from a clean `grpc.io` URL and drop the redirect to `www`.


hsaliak:
what will happen to existing links that point to www.grpc.io?

mehrdada:
Redirected.

PR #543, opened 2017-07-12
Add reference to android codegen

dapengzhang0:


docs/reference/java/generated-code.md
```diff
@@ -212,7 +212,7 @@ Typically the build system handles creation of the gRPC generated code.
 For protobuf-based codegen, you can put your proto files in the `src/main/proto`
 and `src/test/proto` directories along with an appropriate plugin.
 
-For protobuf-based codegen integrated with the Maven build system, you can use
+For protobuf-based codegen integrated with the Maven build system, non-android developers can use
```
mehrdada: Since this is a reference, not a guide, I think a better phrasing would be something like:

`A typical Maven rule for generating gRPC and Protocol Buffers code would look like the following:`

I don't think designating "non-Android" explicitly would be necessary.  However, if you want to add that designation, please uppercase A in android.

PR #571, opened 2017-09-11
Add grpc_2_0 flag to example protoc commands

nathanielmanistaatgoogle:


docs/quickstart/python.md
```diff
@@ -167,7 +167,7 @@ service definition.
 From the `examples/python/helloworld` directory, run:
 
 ```sh
-$ python -m grpc_tools.protoc -I../../protos --python_out=. --grpc_python_out=. ../../protos/helloworld.proto
+$ python -m grpc_tools.protoc -I../../protos --python_out=. --grpc_python_out=grpc_2_0:. ../../protos/helloworld.proto
```
mehrdada: Do you have to run `protoc` separately with `grpc_2_0` to get the non-gRPC code elements? If yes, we need to amend the instructions with a direct `protoc` invocation.

nathanielmanistaatgoogle: No, this is the only command that has to be run. Double-check my work? Locally delete the `_pb2.py` and `_pb2_grpc.py` files, run this command, and verify that the files are wholly regenerated and completely ready for use?

mehrdada: Will do now. Give me a couple minutes.

ctiller:
Curious as to what _2_0 signifies... are we at a major API rev?

nathanielmanistaatgoogle:
@ctiller: the name indicates "behavior that we intend to be the default in 2.0". So it's forward-looking; [I had the same question when it was introduced](https://github.com/grpc/grpc/pull/8920#discussion_r90516569). The flag isn't yet widely used by parties other than ourselves; if you see a problem now with the flag name or intention, now's a good time to tell us to freeze this progress and go a different way.

nathanielmanistaatgoogle:
Very happily abandoning this change. :-)

PR #574, opened 2017-09-22
Add a link to TiDB which also uses gPRC

queenypingcap:
TiDB is also using gRPC, thus adding the link to it.


queenypingcap:
I signed it!

PR #575, opened 2017-09-22
Fix RouteGuideServer.java and RouteGuideClient.java path

MrYawe:


MrYawe:
I signed it!

ericgribkoff:
Good catch, thanks!

PR #583, opened 2017-10-17
Add Python version prerequisite

nathanielmanistaatgoogle:


nathanielmanistaatgoogle:
D'oh! Of course. Fixed.

@fengli79: is a fix like this what you had in mind?

PR #585, opened 2017-10-23
Fix link to Ruby docs

bmarkons:
Seems like Ruby docs are accessible at HTTP instead of HTTPS -- same as here:
https://grpc.io/docs/

🙇


bmarkons:
I have signed it.

PR #591, opened 2017-11-01
Add -u flag to go get grpc

seblegall:
Some updates have to be done on the http2 package when trying to compile grpc. Since the compilation fail if http2 package is not up-to-date, the `update` option should be passed to go get.


seblegall:
I signed it!

hsaliak:
Thanks for the PR!

PR #593, opened 2017-11-10
gRPC PHP: update quickstart and tutorials

ZhouyihaiDing:
1. To solve the issue https://github.com/grpc/grpc/issues/12481 alone with https://github.com/grpc/grpc/issues/13193 : 
Installation instruction should be consistent between grpc, grpc.io while it needs to notify the compatibility between grpc version and protobuf version.
2. To solve the issue https://github.com/grpc/grpc/issues/12229 : update how to run examples.

For 1:
By referring to the grpc/src/php/readme.md, adding installation guidance missed in grpc.io.
- Update installation commands for Pre-request: php, pecl, composer, phpunit.
- Update install grpc extension by pecl and and add installation method by build from source.
- Update other pre-request related to protobuf: Install protoc, protobuf.so and build grpc_php_plugin.
- Add compatibility table between grpc version and protobuf version(It should keep updating when grpc use higher version of protobuf).
- Re-organize the order.
I only try to make the installation part consistent. Since the usage of github and groc.io is different, differences parts remains there: UnitTest introduction only exist in grpc repo and Greeter Example exists in grpc.io.

For 2:
- Add information about how to generate proto files using grpc_php_plugin manually or helper script.
- Fix in greeter example, node server stub doesn't update service.

### DIFF
--- a/docs/quickstart/php.md
+++ b/docs/quickstart/php.md
@@ -12,120 +12,249 @@ working example.</p>
 
 ### Prerequisites
 
-* `php`: version 5.5 or higher, 7.0 or higher
-* `pecl`: version 1.9 or higher (on Linux)
+* `php` 5.5 or above, 7.0 or above
+* `pecl`
 * `composer`
+* `phpunit` (optional)
 
-## Install on Mac OS X
+**Install PHP and PECL on Ubuntu/Debian:**
+
+For PHP5:
+
+```sh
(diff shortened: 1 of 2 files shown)


ZhouyihaiDing:
For the installation part, maybe there will be one more change later if `phpunit` is installed by composer https://github.com/grpc/grpc/pull/13228

PR #609, opened 2017-12-19
Simplify and clarify the instructions for C++ example

mehrdada:
Fixes https://github.com/grpc/grpc/issues/12422

### DIFF
--- a/docs/quickstart/cpp.md
+++ b/docs/quickstart/cpp.md
@@ -16,33 +16,28 @@ working example.</p>
 
 #### Install gRPC
 
-To install gRPC on your system, follow the instructions to build from source
-[here](https://github.com/grpc/grpc/blob/master/INSTALL.md).
+To install gRPC on your system, follow the [instructions to build from source]
+(https://github.com/grpc/grpc/blob/master/INSTALL.md).
+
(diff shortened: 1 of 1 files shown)


docs/quickstart/cpp.md
```diff
@@ -16,34 +16,29 @@ working example.</p>
...
 
 ```sh
-$ git submodule update --init
-$ cd grpc/third_party/protobuf
-$ ./autogen.sh
-$ ./configure
-$ make
-$ sudo make install
+$ cd third_party/protobuf
+$ make install
```
jtattermusch: I have mixed feelings about this.
I thinks simplifying the docs is useful, but on the other hand, I'm not sure if installing protobuf from grpc's submodule is something worth recommending (it's probably fine for the purposes of the examples)

As a side note, I think calling "make install" is actually an anti-pattern IMHO (there is no way to uninstall and that has bitten many users in the past). I'd be much happier if our example instructions mentioned that "make install" is something you shouldn't do and what are the risks.

mehrdada: The issue is that installing protobuf from the tar.gz file won't register it on pkg-config and will cause our examples to fail

PR #616, opened 2018-01-18
Specify gRPC minimum deployment version

muxi:
Specify gRPC minimum deployment iOS version in quick start.


muxi:
@hsaliak needs a review :)

PR #621, opened 2018-01-26
Create 2018-01-26-grpc-and-deadlines.md

LisaFC:
Grainne's deadline blog post. Set to unpublished for now.

### DIFF
--- a/_posts/2018-01-26-grpc-and-deadlines.md
+++ b/_posts/2018-01-26-grpc-and-deadlines.md
@@ -1,0 +1,139 @@
+---
+layout: post
+title: gRPC and Deadlines
+published: 
+permalink: blog/deadlines
+author: Gráinne Sheerin, Google SRE
+company: Google
+company-link: https://www.google.com
+---
+
+**TL;DR Always set a deadline**. This post explains why we recommend being deliberate about setting deadlines, with useful code snippets to show you how.
+
+<!--more-->
+
+When you use gRPC, the gRPC library takes care of communication, marshalling, unmarshalling, and deadline enforcement. Deadlines allow gRPC clients to specify how long they are willing to wait for an RPC to complete before the RPC is terminated with the error `DEADLINE_EXCEEDED`. By default this deadline is a very large number, dependent on the language implementation. How deadlines are specified is also language-dependent. Some language APIs work in terms of a **deadline**, a fixed point in time by which the RPC should complete. Others use a **timeout**, a duration of time after which the RPC times out.
+
+In general, when you don't set a deadline resources will be held for all in-flight requests, and all requests can potentially reach the maximum timeout. This puts the service at risk of running out of resources like memory and crashing, as well as maximizing the latency of the service.
+
+To avoid this, services should specify the longest default deadline they technically support, and clients should wait until the response is no longer useful to them. For the service this can be as simple as providing a comment in the .proto file. For the client this involves setting useful deadlines.
+
+There is no single answer to "What is a good deadline/timeout value?". Your service might be as simple as the Greeter in our quick start guides, in which case 100 ms would be fine. Your service might be as complex as a globally-distributed and strongly consistent database. The deadline for a client query will be different from how long they should wait for you to drop their table.
+
+So what do you need to consider to make an informed choice of deadline? Factors to take into account include the end to end latency of the whole system, which RPCs are serial, and which can be made in parallel. You should to be able to put numbers on it, even if it's a rough calculation. Engineers need to understand the service and then set a deliberate deadline for the RPCs between clients and servers.
+
+In gRPC, both the client and server make their own independent and local determination about whether the remote procedure call (RPC) was successful. This means their conclusions may not match! An RPC that finished successfully on the server side can fail on the client side. For example, the server can send the response, but the reply can arrive at the client after their deadline has expired. The client will already have terminated with the status error `DEADLINE_EXCEEDED`.
+
+## Setting a deadline
+
+As a client you should always set a deadline for how long you are willing to wait for a reply from the server. Here's an example using the greeting service from our [Quick Start Guides]({{ site.baseurl }}/docs/quickstart/):
+
+**C++**
+
+
+```
+  ClientContext context;
+  time_point deadline = std::chrono::system_clock::now() + 
+      std::chrono::milliseconds(100);
+  context.set_deadline(deadline);
+```
+
+
+**GO LANG**
+
+
+```
+    clientDeadline := time.Now().Add(time.Duration(*deadlineMs) * time.Millisecond)
+    ctx, cancel := context.WithDeadline(ctx, clientDeadline)
+```
+
+
+**JAVA**
+
+
+```
+response = blockingStub.withDeadlineAfter(deadlineMs, TimeUnit.MILLISECONDS).sayHello(request);
+```
+
+
+This sets the deadline to 100ms from when the client RPC is set to when the response is picked up by the client. 
+
+
+## Checking deadlines
+
+On the server side, the server can query to see if a particular RPC is no longer wanted. Before a server starts work on a response it is very important to check if there is still a client waiting for it. This is especially important to do before starting expensive processing.
+
+**C++**
+
+
+```
+if (context->IsCancelled()) {
+  return Status(StatusCode::CANCELLED, "Deadline exceeded or Client cancelled, abandoning.");
+}
+```
+
+
+**GO LANG**
+
+
+```
+if ctx.Err() == context.Canceled {
+	return status.New(codepb.CANCELLED, "Client cancelled, abandoning.")
+```
+
+
+**JAVA**
+
+
+```
+if (Context.current().isCancelled()) {
+  responseObserver.onError(Status.CANCELLED.withDescription("Cancelled by client").asRuntimeException());
+  return;
+}
+```
+
+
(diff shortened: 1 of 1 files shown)


_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+
+In general, when you don't set a deadline resources will be held for all in-flight requests, and all requests can potentially reach the maximum timeout. This puts the service at risk of running out of resources like memory and crashing, as well as maximizing the latency of the service.
+
+To avoid this, services should specify the longest default deadline they technically support, and clients should wait until the response is no longer useful to them. For the service this can be as simple as providing a comment in the .proto file. For the client this involves setting useful deadlines.
+
+There is no single answer to "What is a good deadline/timeout value?". Your service might be as simple as the Greeter in our quick start guides, in which case 100 ms would be fine. Your service might be as complex as a globally-distributed and strongly consistent database. The deadline for a client query will be different from how long they should wait for you to drop their table.
```
hsaliak: Could we please link them to the greeter service in our quickstart? Any one will do.

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+  ClientContext context;
+  time_point deadline = std::chrono::system_clock::now() + 
+      std::chrono::milliseconds(100);
+  context.set_deadline(deadline);
+```
+
+
+**GO LANG**
```
hsaliak: We should just replace this with Go

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+
+**JAVA**
+
+
+```
+response = blockingStub.withDeadlineAfter(deadlineMs, TimeUnit.MILLISECONDS).sayHello(request);
+```
+
```
hsaliak: Should we add an example with Timeouts for completeness? Using python perhaps something like
```
response = stub.Foo(request,  timeout= 0.1)  # timeouts are specified in seconds
```
@nathanielmanistaatgoogle does this sound right

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+clientDeadline := time.Duration(*deadlineMs) * time.Millisecond
+ctx, cancel := context.WithTimeout(ctx, clientDeadline)
+```
+
+
+**JAVA**
+
+
+```
+ @Option(name="--deadline_ms", usage="Deadline in milliseconds.")
```
hsaliak: does this need the google-options package?

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+**GO LANG**
+
+
+```
+    clientDeadline := time.Now().Add(time.Duration(*deadlineMs) * time.Millisecond)
+    ctx, cancel := context.WithDeadline(ctx, clientDeadline)
+```
+
+
+**JAVA**
```
mehrdada: s/JAVA/Java

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+```
+if (context->IsCancelled()) {
+  return Status(StatusCode::CANCELLED, "Deadline exceeded or Client cancelled, abandoning.");
+}
+```
+
+
+**GO LANG**
```
mehrdada: s/GO LANG/Go

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+```
+if ctx.Err() == context.Canceled {
+	return status.New(codepb.CANCELLED, "Client cancelled, abandoning.")
+```
+
+
+**JAVA**
```
mehrdada: s/JAVA/Java

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+DEFINE_int32(deadline_ms, 20*1000, "Deadline in milliseconds.");
+
+  ClientContext context;
+  time_point deadline = std::chrono::system_clock::now() + 
+      std::chrono::milliseconds(FLAGS_deadline_ms);
+  context.set_deadline(deadline);
+```
+
+
+**GO LANG**
```
mehrdada: s/GO LANG/Go

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+var deadlineMs = flag.Int("deadline_ms", 20*1000, "Default deadline in milliseconds.")
+
+clientDeadline := time.Duration(*deadlineMs) * time.Millisecond
+ctx, cancel := context.WithTimeout(ctx, clientDeadline)
+```
+
+
+**JAVA**
```
mehrdada: s/JAVA/Java

hsaliak:
@LisaFC would you be able to address the review comments? It would be nice to get this published.

LisaFC:
Ping?

_posts/2018-01-26-grpc-and-deadlines.md
```diff
@@ -0,0 +1,139 @@
...
+
+So what do you need to consider to make an informed choice of deadline? Factors to take into account include the end to end latency of the whole system, which RPCs are serial, and which can be made in parallel. You should to be able to put numbers on it, even if it's a rough calculation. Engineers need to understand the service and then set a deliberate deadline for the RPCs between clients and servers.
+
+In gRPC, both the client and server make their own independent and local determination about whether the remote procedure call (RPC) was successful. This means their conclusions may not match! An RPC that finished successfully on the server side can fail on the client side. For example, the server can send the response, but the reply can arrive at the client after their deadline has expired. The client will already have terminated with the status error `DEADLINE_EXCEEDED`.
```
hsaliak: What happens in this case? 
It would be nice to propose what the client should do in this case. This leaves the server and the client in an inconsistent state. Is it then best practice to (a) include logic on the server side on requests to ensure that methods are idempotent when possible and when not possible, checking if a state change is allowed, or (b) including client side logic to check if a change has been persisted on the server side when deadlines are in use? /cc @ejona86 
At the moment, this paragraph leaves users with an open question on what to do.

_posts/2018-01-26-grpc-and-deadlines.md
nathanielmanistaatgoogle: I'm mostly ambivalent but leaning yes.

_posts/2018-01-26-grpc-and-deadlines.md
ejona86: I don't know of any OSS google options package. Without an import I couldn't tell what library it was using.

_posts/2018-01-26-grpc-and-deadlines.md
ejona86: This problem isn't unique to deadlines. It can happen for I/O errors as well. Idempotency is good. But in some cases it may need the client to re-evaluate its assumptions (which may retry multiple calls in sequence).

hsaliak: Could we in this case suggest adding a line to address this point. 
eg: "Clients/Servers should check for, and manage this at an application level"

@ejona86 sounds like we really do need an advanced error handling blog post. It would be useful to call out a few of these corner cases and the factors that need to be considered when handling them. :)

LisaFC: That's a good idea. For final clarifications on this post can I add Grainne (who actually wrote the blog post) to this discussion as she may have some thoughts? She's @gsheerin.

gsheerin: Hi (thanks all for the reviews and feedback!)

I like the suggestion "Clients and Servers should check for, and manage this at an application level". 
So that would become...
"In gRPC, both the client and server make their own independent and local determination about whether the remote procedure call (RPC) was successful. This means their conclusions may not match! An RPC that finished successfully on the server side can fail on the client side. For example, the server can send the response, but the reply can arrive at the client after their deadline has expired. The client will already have terminated with the status error `DEADLINE_EXCEEDED`.  This should be checked for and managed at the application level."


Can we also add that we'll cover that discussion, in a future blog post?



Some additional thoughts.
- If you miss a deadline (on the client side) you can retry. It has often lead to cascading failures. We need to think carefully about re-try logic before implementing, aka  "if at first you don't succeed, back off exponentially" -- Dan Sandler, Google Software Engineer.
- If your handling methods that are idempotent, it's easier. Be sensible about the re-try.
- If they aren't idempotent,  it's a decision about how to fail. It's probably better to retry deadline exceeded failures to see if it succeeded before retrying the rpc to apply the change again. 
- If you need to make multiple changes to handle a full client request, it's best to check the deadline left at each stage before attempting to preform more work, on the request on the server side. 

I'd be happy to write more material for posts, and I can help out with the error handling one if that helps.

LisaFC:
Are we good yet?

hsaliak:
Approved, please merge on Monday, so that we can publicize it then. 
Thank you very much Lisa and Grainne!

PR #630, opened 2018-02-22
Correct usage of server streaming RPC calls in Node

eatonphil:
Since https://github.com/grpc/grpc/issues/8954 was fixed, it doesn't seem like the `'end'` event is actually sent. This also adds documentation on errors and error-handling semantics.


thelinuxfoundation:

* User @eatonphil isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

eatonphil:
Signed the CLA.

murgatroid99:
This change is not really correct. In normal operation, an `end` event will be fired at the end of the stream as documented. The change from that linked issue means that if there is an error, then there won't be an end event.

eatonphil:
@murgatroid99 got it, thanks. Will update.

murgatroid99:
OK, this looks more accurate, but I don't see why you want to add the sentence "Errors in the `'data'` callback will not cause the stream to be closed." In general, errors in user code won't affect the stream state, and I don't think it's necessary to call out that specific event.

eatonphil:
I think it's better to be explicit, the way it's set up does not make it clear whether errors in the data callback will cause the stream to be cancelled or not. I only saw what really happens when I tried it. There's no reason not to document that. A different library could choice to await Promises returned by the callback and close the stream if it failed. Either way seemed reasonable to me.

eatonphil:
ping

eatonphil:
Anything more I can do to get this merged?

murgatroid99:
Apparently someone else needs to approve this. @hsaliak can you do that?

PR #638, opened 2018-03-09
Add Go examples to auth guide

enocom:
Fixes https://github.com/grpc/grpc-go/issues/1396
Fixes https://github.com/grpc/grpc.github.io/issues/545

/cc @dfawley

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -177,6 +177,41 @@ languages. The following sections demonstrate how authentication and
 authorization features described above appear in each language: more languages
 are coming soon.
 
+### Go
+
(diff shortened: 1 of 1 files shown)


enocom:
@dfawley I've added server examples, as suggested.

I've also updated the example for connecting to a Google API, since I noticed the [oauth.NewComputeEngine](https://github.com/grpc/grpc-go/blob/90dca43332f6cc944c37e16f32a82c41639e7705/credentials/oauth/oauth.go#L114) function seems to be on the path towards deprecation.

enocom:
😄

PR #641, opened 2018-03-21
try out author link, and remove extra header

carl-mastrangelo:
@LisaFC FYI.

I would love to know if this works, but I have no way to preview the change.

### DIFF
--- a/_posts/2018-03-06-so-you-want-to-optimize-grpc-part-1.md
+++ b/_posts/2018-03-06-so-you-want-to-optimize-grpc-part-1.md
@@ -4,11 +4,10 @@ title: So You Want to Optimize gRPC - Part 1
 published: true
 permalink: blog/optimizing-grpc-part-1
 author: Carl Mastrangelo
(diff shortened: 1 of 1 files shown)


LisaFC:
So just FYI, there are two ways you can preview your change
- Fork the entire repo and enable Github Pages for your master branch - it'll then be served as a clone website with your changes in it. (this does mean there are temporary "fake gRPC sites" around while they're up but it doesn't seem to have confused anyone yet)
- Have local copy of repo and run Jekyll manually to have local server.

Kubernetes moved from Github Pages serving to Netlify serving (free for OS projects) because it offered staging/previews (among other advantages), so that might be something for us to consider too. Switching seems to be relatively trivial, having tried with one of my personal repos.

LisaFC:
Here's the instructions for a local Jekyll server (obviously you don't need to do the bit where you create a new website): https://jekyllrb.com/docs/quickstart/

It does require some Ruby installation but once that's all installed it works fine.

PR #651, opened 2018-04-10
fix go cert example

menghanl:
fixes https://github.com/grpc/grpc-go/issues/1980


menghanl:
Conflict fixed! PTAL.

PR #661, opened 2018-05-07
Add more presentation Videos

jtattermusch:
- move videos from About section to docs/talks
- add a few more presentation videos

The endresult can be checked here: http://jtattermusch.github.io/grpc.github.io/docs/talks/


js/common.js
```diff
@@ -9,8 +9,13 @@ var firstScriptTag = document.getElementsByTagName('script')[0];
 firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
 
 // Array of videoIds
-// The key corresponds to the data attributes in about/index.html
+// The key corresponds to the data attributes in docs/talks/index.html
+// TODO(jtattermusch): is there a way to not require editing this when adding new videos?
 var playerInfoList = [
+    {type: 'yt', key: '5QH67PXNbaI'},
```
jtattermusch: btw, I think requiring this is pretty unfortunate, it makes it very hard to add new videos.

jtattermusch:
@LisaFC  do you have any idea why the "tile layout" is broken in http://jtattermusch.github.io/grpc.github.io/docs/talks/?

jtattermusch:
@LisaFC  friendly ping.

hsaliak:
@LisaFC ping

LisaFC:
Huh, not sure why the tiling is broken, I'll have a look. Have added the content in anyway while we fix the layout....

jtattermusch:
@LisaFC  two videos also fail to load: https://grpc.io/docs/talks/  Not sure what's wrong but I'm pretty sure that wasn't the case back when I created the PR.

PR #662, opened 2018-05-09
small go post fixes

jadekler:
- Missing bracket
- codepb.CANCELED -> status.Canceled

### DIFF
--- a/_posts/2018-02-26-grpc-and-deadlines.md
+++ b/_posts/2018-02-26-grpc-and-deadlines.md
@@ -121,8 +121,7 @@ context.set_deadline(deadline);
 ```go
 var deadlineMs = flag.Int("deadline_ms", 20*1000, "Default deadline in milliseconds.")
 
-clientDeadline := time.Duration(*deadlineMs) * time.Millisecond
-ctx, cancel := context.WithTimeout(ctx, clientDeadline)
+ctx, cancel := context.WithTimeout(ctx, time.Duration(*deadlineMs) * time.Millisecond)
 ```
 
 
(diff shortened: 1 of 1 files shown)


dfawley:
While you're at it, can you also change:

```go
clientDeadline := time.Now().Add(time.Duration(*deadlineMs) * time.Millisecond)
ctx, cancel := context.WithDeadline(ctx, clientDeadline)
```

to

```go
ctx, cancel := context.WithTimeout(ctx, time.Duration(*deadlineMs) * time.Millisecond)
```

jadekler:
Done

_posts/2018-02-26-grpc-and-deadlines.md
```diff
@@ -78,7 +78,8 @@ if (context->IsCancelled()) {
 
 ```go
 if ctx.Err() == context.Canceled {
```
dfawley: More complete would be:

```go
switch ctx.Err() {
  case context.Canceled:
    return status.New(codes.Canceled, "client cancelled, abandoning.")
  case context.DeadlineExceeded:
    return status.New(codes.DeadlineExceeded, "deadline exceeded")
  case nil:
    break
  default:
    return status.New(codes.Unknown, "unknown context error")
}
```

Hmm, maybe we should add a `status.FromContextError(error) *status.Status` or `status.ConvertContextError(error) error` convenience function that does this, so the user's code is:

```go
if err := ctx.Err(); err != nil {
  return status.FromContextError(error).Err()
}
```

Sadly, until 1.6 support is gone, we'll need 2 conditionally compiled files to do this so we can handle both "context" and "x/net/context" for go1.7 and go1.8.. but it's probably worth it.

jadekler:
Right on. I actually don't have contributor access to this repo - could you squash + merge? Or, should I ask @hsaliak for access?

PR #298, opened 2016-08-10
Update Python tutorial to GA API

nathanielmanistaatgoogle:
The only surprise here is that this got past me as long as it did.

@kpayson64 for code; @LisaFC for content.

PR #671, opened 2018-05-29
Update to account for import error in generated python code

dray92:
After generating relevant server/client code for python, the presence of

```python
from google.api import annotations_pb2 as google_dot_api_dot_annotations__pb2
```
causes,

```bash
    from google.api import annotations_pb2 as google_dot_api_dot_annotations__pb2
ImportError: No module named api
```

`pip install grpcio-tools` only brings in `google/protobuf` module.

Installing `googleapis-common-protos` causes the required files, specifically, 
`google/api/annotations_pb2.py` to be pulled in, thus resolving the import error.

There are examples like [python-grpc-hack.sh](https://github.com/kasey/grpc-example/blob/master/installers/python-grpc-hack.sh) out there, which is basically hacking
around the absence of the required module. I have not yet tried the Java codegen yet
but till the relevant protoc changes are pushed upstream, this change seems to be
the path of least resistance. 
What are your thoughts around this?

Related issue: [#298](https://github.com/grpc-ecosystem/grpc-gateway/issues/298).

Signed-off-by: Debosmit Ray <debo@uber.com>

### DIFF
--- a/docs/quickstart/python.md
+++ b/docs/quickstart/python.md
@@ -71,7 +71,7 @@ tutorials and your own projects.
 To install gRPC tools, run:
 
 ```sh
-$ python -m pip install grpcio-tools
+$ python -m pip install grpcio-tools googleapis-common-protos
 ```
 
 ## Download the example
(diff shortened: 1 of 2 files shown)


hsaliak:
@debosmit would you be OK with signing the CLA? without it, we cannot merge the PR.

hsaliak:
Thank you for the contribution!

PR #684, opened 2018-06-19
kotlin post: fixed size image, update title

zpencer:
Add property to make the image not take up the entire screen (hopefully).


hsaliak:
Can we retitle to - gRPC ❤ Kotlin

PR #685, opened 2018-06-19
undo image resize attempt

zpencer:
This did not do the trick


hsaliak:
doesnt this revert the image back to the big image?

PR #697, opened 2018-06-27
Remove backslash as it is not rendered correctly

dapengzhang0:
also add missing `throws` in one of the code snippets.


dapengzhang0:
@carl-mastrangelo

PR #706, opened 2018-07-11
Make overview links helpful

rvandegrift:
Currently, the Overview links go to the docs overview page, not the grpc overview.  This isn't very helpful to a new reader, and it took me a minute to find the actual grpc overview from the index page.

If this is accepted a similar change to other languages would probably be beneficial.


carl-mastrangelo:
@rvandegrift LGTM

PR #728, opened 2018-08-21
Update C# quickstart and tutorial

jtattermusch:
Fixes https://github.com/grpc/grpc/issues/16397

We should hold merging until v1.15.0 is released (~about two weeks from now), because the changes to C# examples from https://github.com/grpc/grpc/pull/16144  won't be propagated to the release branch until then.


jtattermusch:
This will need updating after https://github.com/grpc/grpc/pull/16518 is merged.

PR #766, opened 2018-10-23
gRPC guarantees the order of streaming messages

alexlo03:
Documented well in the bi-di case, but not explicitly spelled out in the other streaming cases.

### DIFF
--- a/docs/guides/concepts.md
+++ b/docs/guides/concepts.md
@@ -55,7 +55,8 @@ rpc SayHello(HelloRequest) returns (HelloResponse){
 
 - Server streaming RPCs where the client sends a request to the server and gets
   a stream to read a sequence of messages back. The client reads from the
-  returned stream until there are no more messages.
+  returned stream until there are no more messages. gRPC guarantees message
+  ordering.
 
 ```
 rpc LotsOfReplies(HelloRequest) returns (stream HelloResponse){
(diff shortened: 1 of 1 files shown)


alexlo03:
CLA signed

docs/guides/concepts.md
```diff
@@ -55,7 +55,8 @@ rpc SayHello(HelloRequest) returns (HelloResponse){
 
 - Server streaming RPCs where the client sends a request to the server and gets
   a stream to read a sequence of messages back. The client reads from the
-  returned stream until there are no more messages.
+  returned stream until there are no more messages. gRPC guarantees message
+  ordering.
```
carl-mastrangelo: nit: gRPC guarantees message order for a given stream.   The messages from different streams can be reordered.

alexlo03: Away from a github enabled machine - would either of these wordings be preferred? 
A) "gRPC guarantees message ordering within an individual RPC call." 
B) "gRPC guarantees message ordering within an individual stream."

carl-mastrangelo: Let's say call, since stream is a transport specific thing.

alexlo03:
@carl-mastrangelo Updated per our conv

carl-mastrangelo:
@alexlo03 Merged, thanks!

PR #785, opened 2018-12-03
Revert "Update to account for import error in generated python code"

ericgribkoff:
This reverts commit d9c953d22bb59f07d01e90053fee1673252e582c.

cc @hsaliak I'm not sure why this PR was merged. Unless I'm missing something, the reported error was because the *user's* protobuf file (not grpcio-tools itself) had a dependency on googleapis-common-protos. They should add this to their project's dependencies, it's not something required to use grpcio-tools or our quickstart or tutorial code.

### DIFF
--- a/docs/quickstart/python.md
+++ b/docs/quickstart/python.md
@@ -71,7 +71,7 @@ tutorials and your own projects.
 To install gRPC tools, run:
 
 ```sh
-$ python -m pip install grpcio-tools googleapis-common-protos
+$ python -m pip install grpcio-tools
 ```
 
 ## Download the example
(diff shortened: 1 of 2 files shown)


hsaliak:
Agreed - googleapis-common-protos is not something that was needed for gRPC.
Apologies for missing this in the earlier review and thank you for catching it :+1:

lidizheng:
Do we need to refer to `googleapis-common-protos` at some place of our tutorial? I guess it need proper documentation for that package, so people can use it properly.

1. How to import it?
2. What is the content?
3. How to find documentation for each proto?

PR #815, opened 2019-01-09
add k8s load balancing with linkerd blog post

wmorgan:
Here's a PR for adding a gRPC load balancing on K8s post, similar to the one published on the K8s blog.

I've added it under the 2019-01-15 date, but that was arbitrary. Happy to change that.

I also ended up tweaking the SCSS to not autoscale images on the post permalink pages, which was causing things to look very blurry on the post permalink, but I'm not 100% sure that's the right way to do it.

There's a preview of this https://wmorgan.github.io/grpc.github.io/blog/.

Feedback welcome!


carl-mastrangelo:
cc @hsaliak

_posts/2019-01-15-grpc-load-balancing-on-k8s-with-linkerd.md
```diff
@@ -0,0 +1,176 @@
...
+benefits, including dramatically lower (de)serialization costs, automatic type
+checking, formalized APIs, and less TCP management overhead.
+
+However, gRPC also breaks the standard connection-level load balancing,
+including what's provided by Kubernetes. This is because gRPC is built on
+HTTP/2, and HTTP/2 is designed to have a single long-lived TCP connection,
+across which all requests are *multiplexed*&mdash;meaning multiple requests can be
+active on the same connection at any point in time. Normally, this is great, as
+it reduces the overhead of connection management. However, it also means that
+(as you might imagine) connection-level balancing isn't very useful. Once the
```
carl-mastrangelo: I'm not sure I agree with this.  gRPC comes with two client-side load balancers by default: Pick First and Round Robin.  Pick first has the behavior you describe, but Round Robin does cycle requests across each connection.

wmorgan: AIUI you're talking about client request-level balancing, but this paragraph is about connection-level balancing, which is something that K8s provides.

carl-mastrangelo: I think we have different definitions for connection-level load balancing.   For me, connection-level load balancing means sending new connections to different backends and pinning them there.  It's some that that a proxy would do, not a gRPC client though.  (since a gRPC client would only create a single connection to any given backend).  

What do you mean by "However, gRPC also breaks the standard connection-level load balancing" ?

wmorgan: That's my definition of connection-level balancing as well.

K8s provides connection-level balancing, and it does that via a proxy (kubeproxy). My point in this paragraph is that this K8s-provided connection-level load balancing, which works fine with HTTP/1.1, stops working with gRPC. If you make HTTP/1.1 calls to a K8s "Service", everything will Just Work -- kubeproxy will distribute load across all the backends for that Service. But if you simply replace HTTP/1.1 with gRPC, this breaks, as your calls all get pinned to an individual backend. So you need to do stuff with the gRPC client, or to use an L7 proxy, or something other option--all of which I go into later in the post.

If there's a better way of expressing the above, I'm all ears.

srini100: Hi @wmorgan, the reason why gRPC breaks connection level load balancing is not because it uses  HTTP/2. I think that statement needs to be clarified. Out of the box, gRPC comes with pick-first and round-robin load balancing options. In round-robin, a single TCP connection is opened to each backend and requests are load balanced on these connections. This is effectively same as what a proxy would do but with all the additional benefits of HTTP/2. So far so good. The load balancing breaks when new backends are added. Since gRPC client is relying on DNS and it resolves DNS only when connections break, newly added backends are not added to the round robin load balancing. This is the default behavior. To solve this issue, one can implement their own gRPCLB look-aside balancer. gRPC clients can talk to such balancers and get notified of changing backends. Another option is to use a custom resolver instead of DNS that can push changes to gRPC client. Resolver API is public in Java and Go and is there is work in progress in C++. One can also control the connection lifetime on gRPC servers to mitigate issues due to long lived connections and load balancing. See [this](https://github.com/grpc/proposal/blob/ce3468ade738300c94d6b4437ff05cdcdab7f465/A9-server-side-conn-mgt.md).

This is probably too much detail and digressing for your blog so you could probably add a short summary of what gRPC can do. Something like "gRPC has built-in round robin load balancing of requests over single TCP connection to each backend but it requires custom resolver or a look-aside gRPCLB server to account for changing pool of backends. gRPC servers can be configured to control connection lifetime to mitigate the issues due to long lasting connections and load balancing." How does this sound? Thanks.

thisisnotapril:
@wmorgan hi there! we are in the process of moving the gRPC site over to Hugo, and in that process we will need to close PRs and probably reopen them afterwards. That OK with you?

wmorgan:
No worries, I'm not pursuing this any more so I've just closed it.

PR #826, opened 2019-01-23
Fix link to detailed C++ performance dashboard.

jtattermusch:
The old link leads to a personal dashboard of a team member (and there's no data).


jtattermusch:
CC @mhaidrygoog @hsaliak

PR #832, opened 2019-02-17
Add links to grpc-spring-boot-starter

ST-DDT:
There is the [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter) library that connects gRPC with spring. IMO it would be useful, if we could link to that project to allow new users of this framework to discover its integration with the spring framework.

I'm not 100% sure, whether I placed the links at the correct location or whether I should add some more documentation for the actual integration there. I didn't include it for now, because it is already present in the linked project.

**See also:**
* https://github.com/grpc/grpc-java/pull/5368
* https://github.com/spring-projects/spring-framework/issues/20905
* https://github.com/spring-cloud/spring-cloud-sleuth/pull/1139#issuecomment-444302714

**Alternatives considered:**
* Create an official gRPC project for spring integration

**Disclosure:** I'm one of the core maintainers of [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter).

### DIFF
--- a/docs/quickstart/java.md
+++ b/docs/quickstart/java.md
@@ -201,6 +201,8 @@ Just like we did before, from the `examples` directory:
 - Work through a more detailed tutorial in [gRPC Basics: Java][]
 - Explore the gRPC Java core API in its [reference
   documentation](/grpc-java/javadoc/)
+- Try gRPC with [Spring](https://spring.io/) using the unofficial
+  [grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter) library
 
 [gRPC Basics: Java]:../tutorials/basic/java.html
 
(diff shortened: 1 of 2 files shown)


carl-mastrangelo:
Hi,

There is a repo called [Awesome gRPC](https://github.com/grpc-ecosystem/awesome-grpc) which we have been putting projects that make gRPC easier to use.  (and the spring code is linked there already!).  I would feel more comfortable linking to the Java heading of that doc, rather than mentioning the project directly.

@hsaliak  WDYT?

hsaliak:
Awesome gRPC is the right place to add the link into. I'd recommend adding it there.. if the project itself wants to move to the grpc-ecosystem org, we can consider that too!

PR #833, opened 2019-02-21
Update go client streaming example

Teddy-Schmitz:
When the server returns an error during a client-streaming RPC the error on the client side will be `io.EOF` in order to get the real message back from the server you need to call stream.CloseAndRecv() and then log that.

### DIFF
--- a/docs/tutorials/basic/go.md
+++ b/docs/tutorials/basic/go.md
@@ -510,7 +510,9 @@ if err != nil {
 }
 for _, point := range points {
 	if err := stream.Send(point); err != nil {
-		_, err := stream.CloseAndRecv()
+		if err == io.EOF {
+			break
+		}
 		log.Fatalf("%v.Send(%v) = %v", stream, point, err)
 	}
 }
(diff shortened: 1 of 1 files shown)


docs/tutorials/basic/go.md
```diff
@@ -510,6 +510,7 @@ if err != nil {
 }
 for _, point := range points {
 	if err := stream.Send(point); err != nil {
+		_, err := stream.CloseAndRecv()
```
menghanl: How about `break` if err is `io.EOF`, so `CloseAndRecv` after this for loop will get the error?

Teddy-Schmitz: sure that makes it more in-line with the server streaming example, updated

menghanl:
@srini100 
It seems I don't have permission to approve and merge. Can you help me to merge this? Thanks!

carl-mastrangelo:
@Teddy-Schmitz merged, thanks!

menghanl:
Thank you @carl-mastrangelo

PR #834, opened 2019-02-24
Fix go client bidirectional example

thegrumpylion:
In the example of handling bidirectional streaming in the client side,
stream.CloseSend() was called before <-waitc which makes no sense. It
should be the other way around

### DIFF
--- a/docs/tutorials/basic/go.md
+++ b/docs/tutorials/basic/go.md
@@ -558,8 +558,8 @@ for _, note := range notes {
 		log.Fatalf("Failed to send a note: %v", err)
 	}
 }
-stream.CloseSend()
 <-waitc
+stream.CloseSend()
 ```
 
 The syntax for reading and writing here is very similar to our client-side


thelinuxfoundation:

* User @thegrumpylion isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

dfawley:
`stream.CloseSend()` lets the server know that the client is done sending messages.  `waitc` is closed when the server successfully terminates the stream (the client detects this by getting `io.EOF` from `Recv()`).  It's normal for servers to wait until the client does a `CloseSend()` before ending the RPC.

If you check out the sample server code in our repo [here](https://github.com/grpc/grpc-go/blob/871b88ce2e8b9d8e1304740aa55c568cde7641ee/examples/route_guide/server/server.go#L129-L131), you will notice that this is exactly what our server does.  If we were to reverse the lines as proposed, the client would never indicate it was done sending messages, the server would wait forever, and the RPC and the client would never terminate (until the deadline is reached).

PR #872, opened 2021-04-08
Add favicon.ico

sergiitk:
🐕 🥞


ejona86:
While this repository is dead, we still do use gh-pages for some language-specific hosting, like javadoc. So this change can still make sense even though it appears there is no content.

ejona86 (approved):

sergiitk:
Ref https://github.com/grpc/grpc-java/pull/8062

